Issue #10, opened 2021-01-16
Feature request: (maybe) rename `.toMatchTypeOf`


papb:
Hi @mmkal! In mmkal/ts#30 I asked for clarification on `.toMatchTypeOf`, because its meaning wasn't clear to me from the name alone. You've updated the documentation and it is clear after reading it, but now that some time passed I had to check again because I again wasn't sure.

_Original request_

I'd like to request the addition of `.toBeSubtypeOf` as an alias to `.toMatchTypeOf`, and also request the addition of `.toBeSupertypeOf` while you're at it. What do you think?

mmkal:
I'd like to separate out the idea of `.toBeSuperTypeOf` into a separate issue, since I'm not sure I think the added cognitive overhead is worth it. If you want to create a new issue with some use cases, feel free though.

Options as I see them:

(Also @papb I've invited you as a collaborator, which (maybe) will allow you to edit this comment, so if you think of another option, you can add here)

1. `expectTypeOf<Apple>().toMatchTypeOf<Fruit>()`:
   1. Pro: Incumbent
   2. Pro: Similar to jest `expect({a: 1, b: 2}).toMatchObject({a: 1})`
   3. Con: Unclear what it means for non-object types
   4. Con: Unclear generally what "match" means in a type context?
   5. Con: Kinda ambiguous even when you've really thought about it a lot
2. `expectTypeOf<Apple>().toBeSubTypeOf<Fruit>()`:
   1. Pro: Pretty clear if you're familiar with types
   2. Pro: Also clear for non-object types? `expectTypeOf<'apple'>().toBeSubTypeOf<'apple' | 'banana'>()`
   3. Cons: None?
3. `expectTypeOf<Apple>().toBeAssignableToTypeOf<Fruit>()` (suggested in #12)
   1. Pro: Unambiguous - corresponds directly to the real typescript feature
   2. Con: [IMO] not all that intuitive. Nobody's assigning anything, really, so it forces the user to imagine a hypothetical scenario

I think 2 probably is the best, _aside_ from toMatchTypeOf already existing and being in use quite a lot. Maybe it could be kept around as an alias though, although I also don't love doing that.

trevorade:
Yeah, I also like `2.` above. Assignable is definitely not a good word to use because, as far as I'm aware, `expect-type` does not actually check for assignability anywhere and it's impossible for it to check for "not assignable"

[...]

PR #12, opened 2022-10-18
Update Extends to not distribute over union types

trevorade:
Was attempting to use `expectTypeOf` with unioned types and I was surprised that `expectTypeOf` was always mismatching.

See [this repro](https://www.typescriptlang.org/play?#code/C4TwDgpgBAogHsCA7AJgZwDwBkA0UBKAfFALxRZQQLLoFQD8UwATgK7QBcUAZgIYA2aCAG4AUKNCQoASVQBLZhADGwfiFKxqqTGhZykAcwA+SVgFsARhGZ5Tl60YtyD+4ITEB6D1B9QAevQS4NAAIgrKqupkusz6xnZWzJRatAkOTi5IwAxMbJw8AkKe3r4B4kFSAGJycBAo8Ija2HhEGgDaWAC6yY20bfjdjCzsUFx8giLiktDVtSiyKOEqahqzdQ00OnqGJuaJtnvpzq7uol6+-oFAA) showcasing the problems of union types that get stuck in `Extends`.

Note: This only impacts the LHS of the extends.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,13 @@ expectTypeOf(1).not.toBeUndefined()
 expectTypeOf(1).not.toBeNullable()
 ```
 
+Detect assignability of unioned types:
+
+```typescript
+expectTypeOf<number>().toMatchTypeOf<string | number>()
+expectTypeOf<string | number>().not.toMatchTypeOf<number>()
+```
+
 Use `.extract` and `.exclude` to narrow down complex union types:
 
 ```typescript
--- a/src/index.ts
+++ b/src/index.ts
@@ -80,7 +80,7 @@ type ReadonlyEquivalent<X, Y> = Extends<
   (<T>() => T extends Y ? true : false)
 >
 
-export type Extends<L, R> = IsNever<L> extends true ? IsNever<R> : L extends R ? true : false
+export type Extends<L, R> = IsNever<L> extends true ? IsNever<R> : [L] extends [R] ? true : false
 export type StrictExtends<L, R> = Extends<DeepBrand<L>, DeepBrand<R>>
 
 export type Equal<Left, Right> = And<[StrictExtends<Left, Right>, StrictExtends<Right, Left>]>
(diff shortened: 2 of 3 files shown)


papb:
Did you try `StrictExtends`?

trevorade:
Taking a look at the failing checks now...

trevorade:
That should be good to go now. Updated the `README.md` and made the linter happy.

papb:
Ahh I think I get it now: without your fix, `expectTypeOf<string | number>().not.toMatchTypeOf<number>()` does not compile, while it should. Am I right?

trevorade:
> Ahh I think I get it now: without your fix, `expectTypeOf<string | number>().not.toMatchTypeOf<number>()` does not compile, while it should. Am I right?

Correct!

trevorade:
@mmkal Can you please take a look at this PR? It's pretty straightforward and solves a bug where the added tests in this PR do not work without the change in `index.ts`

mmkal:
@trevorade I ran the tests. I agree this is probably better since it errs on the side of strictness, but it's got me thinking about whether `toMatchTypeOf` is really suitable for this kind of type. It was originally conceived as an equivalent of jest's `toMatchObject` but I guess it doesn't have the "object" in the name.

Interestingly this works as expected already:

```ts
expectTypeOf<{userId: string | number}>().not.toMatchTypeOf<{userId: number}>()
```

I'm still ok with this because it's less disruptive than making a breaking change to only support objects.

Also, linking this issue since my brain isn't finding `.toMatchTypeOf` all that intuitive in this case: https://github.com/mmkal/expect-type/issues/10

trevorade:
Awesome. Anything else you'd like here?

mmkal:
Nope, other than any feedback you might have on what the naming _should_ be. Thank you for the PR!

trevorade:
> Nope, other than any feedback you might have on what the naming _should_ be. Thank you for the PR!

I'd probably rename `toMatchTypeOf` to `isAssignableTo`.

That's easier to intuit about IMO.

trevorade:
Or `toBeAssignableToTypeOf` I suppose to keep the naming convention.

mmkal:
https://github.com/mmkal/expect-type/releases/tag/v0.15.0

PR #14, opened 2022-10-23
port throw types PR from mmkal/ts#152

mmkal:
Update Oct 2023:

I'm going to close this since #16 improves error messages and the [TypeScript PR was closed](https://github.com/microsoft/TypeScript/pull/40468#issuecomment-1670705761).

Copy-paste of https://github.com/mmkal/ts/pull/152

Note - since this is a copy paste of a mod of an old version, a few features that have gone in since then may be regressed. This isn't going to go in  anyway unless throw-types are added to TypeScript officially though.

---

original pr body

depends on https://github.com/microsoft/TypeScript/pull/40468

Error messages before:
image

Error messages after:
image

PR #16, opened 2022-10-23
Improve CLI error messages

### LINKED ISSUE #4: [expect-type] expectTypeOf(this).to{Equal,Match}TypeOf(this) fails
Here is a [TS Playground link](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgUwB5mQYxgFQJ4YDyAZnAL5zFQQhwDkaG2AtDAcnQNwBQ3j08NhjgBhCADsYaGCQA8ASXEBnGAENxmZAD44AXm5w4iles0pUU8QBMlcABQA6J6qgBzJQC4445ADdkUADaALoAlHo6AK7iANbiEADu4gaGcAD83n4BKYZexmoayLzcmAA2qkq2AIKIKap6cACMPGS8jFi47HJiktJyVVpadqEOMBAAogCOkaql+ETEsgPDPO3Y88jdElIWcsMR9kjEEBBe4pEgAEYB5KGDw6MT07Mbcj7+UEOhPHzoAnBlCq2ABCtUMlwaACJLpCfoZMBIVFBIthoPsEDk4KVkPBUA0YAALYBKHipcxMToLWRCZAQUioL6PACyqhgmAJr0WNLpcAZKxSrRSrhxlBO6Mx2Nx+KJJMxa0pm2IdkJxJGYymMzmXSVKqU3zl6A6nOVMrVEBZbI52pNqtJqXlnOp7B5fLNFvZju59K+dsMUBxkSg4nJRu1sh6OxkXOd3sZY3dVqpXt5PoF3DIQA) and [related StackOverflow post](https://stackoverflow.com/questions/73602871/how-do-you-conditionally-type-between-a-function-and-an-instance-of-a-class)

And the sample code:

```ts
export class B {
  b = "b";
  
  get foo() {
    expectTypeOf(this).toEqualTypeOf(this); // errors
    expectTypeOf(this).toMatchTypeOf(this); // errors
  }
}
```

Hovering over the `toEqualTypeOf` call provides this information:
```
(property) ExpectTypeOf<this, true>.toEqualTypeOf: <unknown>(...MISMATCH: MismatchArgs<And<[Extends<DeepBrand<this>, {
    type: "special";
    never: false;
    any: false;
    unknown: true;
}>, {
    type: "special";
    never: false;
    any: false;
    unknown: true;
} extends DeepBrand<...> ? true : false]>, true>) => true (+1 overload)

--- 

No overload matches this call.
  Argument of type '[this]' is not assignable to parameter of type 
    'MismatchArgs<
      And<
        [
          Extends<
            DeepBrand<this>, 
            { 
              type: "special"; 
              never: false; 
              any: false; 
              unknown: true; 
            }
          >, 
          { 
            type: "special"; 
            never: false; 
            any: false; 
            unknown: true; 
          } extends DeepBrand<...> 
            ? true 
            : false
        ]
      >, 
      true
    >'.
  
Argument of type '[]' is not assignable to parameter of type   
  'MismatchArgs<
    And<
      [
        Extends<DeepBrand<this>, DeepBrand<this>>, 
        Extends<DeepBrand<this>, DeepBrand<this>>
      ]
    >, 
    true
  >'.(2769)
```

### LINKED ISSUE #6: [expect-type]: Allow negation of `toBeCallableWith`
I've noticed it's impossible to check for a function to **not** be callable with arguments

https://github.com/mmkal/ts/blob/main/packages/expect-type/src/index.ts#L111

In case of negated assertion, the check property is `never`.

So. I've tweaked around and I think the issue is that typing errors are generated during the last call, where type parameters are merged to generate a `MISMATCH`, spread as a rest parameter to generate type that can never be instantiated properly. But the current format of `toBeCallableWith` already takes a rest parameter for args.

So, here is the proposal: create a new method that takes arguments not as a rest tuple, but as a single tuple. Something like this:

```ts
export interface ExpectTypeOf<Actual, B extends boolean> {
    toBeCallableWithArgs: <T extends any[] | [never]>(args: T, ...MISMATCH: MismatchArgs<Extends<T, Params<Actual>>, B>) => true;
}
```

Thoughts ?

### LINKED ISSUE #17: Unreadable error messages
I introduced this package some weeks ago in the storybook monorepo and I love it.

However I get complains that the error messages are unreadable:

843F203A-565E-4401-8790-2BC41A429AFA

Which I agree on, but I learned myself some tricks to debug that. I do something like this:

```ts
expectTypeOf(meta).toEqualTypeOf<
  ComponentAnnotations<
    VueFramework,
    {
      readonly disabled: boolean;
      readonly label: string;
      onMyChangeEvent?: (id: number) => any;
      onMyClickEvent?: (id: number) => any;
    }
  >
>();

// debugging the error
const _: ComponentAnnotations<
  VueFramework,
  {
    readonly disabled: boolean;
    readonly label: string;
    onMyChangeEvent?: (id: number) => any;
    onMyClickEvent?: (id: number) => any;
  }
> = meta;

const __: typeof meta = _;
```

However, we can not expect that from any (external) contributor.

Would it be possible to make those errors more readable?

### LINKED ISSUE #32: `toBeBoolean` behaves differently from `toEqualTypeOf<boolean>`
[Playground link](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgUwB5mQYxgFQJ4YDyAZnAL5zFQQhwDkaG2AtDAcnQNwBQ3ANsnioAhAC44AQwB2eHoyy52JABSoAlADoYEAELIdECAOnK1c9AvxFiqzdoCiARwCuEvleQkAPACNDxqQA+U04gA).

```ts

let x!: any;
expectTypeOf(x).toBeBoolean();
expectTypeOf(x).toEqualTypeOf<boolean>();
```

The first `expectTypeOf` passes (which it should not) but the second errs (correctly). They should both err.

P.S. The same is true for other such helpers (like `toBeArray`, etc).

### LINKED ISSUE #36: Expose "toExtend" functions
Alongside the existing `ExpectTypeOf` methods, a method to check for successful extension of types would be helpful. Perhaps a strict/branded version as well.

Proposed syntax:
```
expectTypeOf<5>().toExtend<number>();
expectTypeOf<unknown>().toExtend<any>();
expectTypeOf('abc').toExtend([{}]); // Typescript error

expectTypeOf<5>().toStrictExtend<number>();
expectTypeOf<unknown>().toStrictExtend<any>(); // Typescript error
expectTypeOf('abc').toStrictExtend([{}]); // Typescript error
```

I would be happy to submit a PR for this. Especially since the `Extends` and `StrictExtends` types are already implemented, just not directly exposed in the main method.

My current workaround is:
```
expectTypeOf<
     StrictExtends<Foo, Bar>
>().toBeEqual<true>();
```

Admittedly the grammar is a bit different (most methods start with `toBe`) so open to a renaming to match. `toBeExtending`...?

### LINKED ISSUE #37: `StrictExtends` is laxer than `Extends`
Hi, I was trying to use the `Extends` types (see https://github.com/mmkal/expect-type/issues/36) and ran into some unexpected behavior with `StrictExtends`.

My _understanding_ is strict extension should be a subset of normal extension. That is to say, there should be no cases where a type _strict extends_ another type, but does not traditionally extend it. If that is not the case, I would greatly appreciate some explanation on what these edge cases are.

Example:
```
type A = (args: string) => 5;
type B = (args: string | number) => number;

type doesExtend = A extends B ? true : false;
// doesExtend = false
// A cannot extend B because the parameters are too strict

expectTypeOf<
    Extends<A, B>
>().toEqualTypeOf<false>(); // Correct / expected


expectTypeOf<
    StrictExtends<A, B>
>().toEqualTypeOf<true>(); // Incorrect / unexpected?
```

My understanding is that is caused by this line: https://github.com/mmkal/expect-type/blob/main/src/index.ts#L43C11
It spreads the params + return equally.

`extends` of methods is a bit counter intuitive.

In order for method A to extend B:
* A's return type must be "stricter" than B's (`ReturnType<A> extends ReturnType<B>`)
  * This is handled correctly (as far as I can tell)
* A's _parameters_ must be "laxer" than B's (`Parameters<B> extends Parameters<A>`)
  * This is not handled correctly

### LINKED ISSUE #38: expectTypeOf allows an incorrect any value instead of the specified type
When I use `expectTypeOf`, I expect it to fail when I pass `any` as input. For example, `toBeString()` should strictly only allow strings and not any other type...including `any`.

```ts
expectTypeOf(1).toBeString(); //  Fails 👍
expectTypeOf(1 as any).toBeString(); // Passes 👎
```

This is particular useful in cases when your types reason about `any`, I need to strictly know that a type will fail if its `any` or not. For example:

```ts
type IsAny<T> = 0 extends 1 & T ? true : false;

export type CustomType<T> = IsAny<T> extends true ? number: any;
```
If I use `CustomType` somewhere, I'd like to test `expectTypeOf(value).toBeNumber();` and have it fail if it's `any` when I don't expect it to be.

This issue was originally opened at https://github.com/vitest-dev/vitest/issues/4205.

mmkal:
Fixes #17
Closes #36 (by documenting that `toMatchTypeOf` ~= `toExtend`)
Closes #37 (by documenting that helper types are not part of the API surface)
Closes #32
Closes #4
Closes #6 (by documenting that you _can't_ use `.not.toBeCallableWith(...)`)
Closes #38

Use generics to give better error messages than "Arguments for the rest parameter 'MISMATCH' were not provided" for most `toEqualTypeOf` cases and many `toMatchTypeOf` cases. This trades off some implementation complexity for better end-user error messages. Essentially, write a special `<Expected extends ...>` constraint for each overload of each method, which does some crazy conditional logic, which boil down to:

- `<Expected extends Actual>` for `toEqualTypeOf` (when we aren't in a `.not` context)
- `<Expected extends Satisfies<Actual>>` for `toMatchTypeOf`

Anyone interested, have a look at the snapshot updates in `errors.test.ts.snap` to see what the real world difference is.

Each of these constraints apply only when we know it's going to "fail" - i.e. `Satisfies<...>` is a fairly meaningless helper type that is used to try to show errors at the type-constraint level rather than the `...MISMATCH: MismatchArgs<...>` level which won't give good error messages.

When using `.not`, the constraint just becomes `extends unknown`, and you'll have to squint as before.

See also: #14 for the better long-term solution, _if_ the TypeScript team decide to merge the throw types PR.
See also: #13 for a hopefully-complementary improvement to the information on hover, which will improve the cases this doesn't cover.

TODO:

- [x] See if the `expectTypeOf({a: 1}).toMatchTypeOf({a: 'one'})` case can also be improved.
- [x] Document. The constraints are a bit different to what most users would be used to, so it's worth highlighting the best way to read error messages and clarify when they might default to "Arguments for the rest parameter 'MISMATCH' were not provided"

Note: I have publish v0.17.0-1 based on this PR and will hopefully be able to use [that version in vitest](https://github.com/vitest-dev/vitest/pull/4206) as a test before merging.


trevorade:
Nice! Happy to have some better error messages in the command line.

I'll take a closer look tomorrow.

mmkal:
Update: looks like it's possible to get rid of `...MISMATCH` completely. But some more tests/handling of edge cases around `never`/`any` are needed. Might also want to automatically snapshot test _all_ the `// @ts-expect-error` cases in the main test file. Will push commits for that soon.

README.md
```diff
@@ -87,6 +87,15 @@ To allow for extra properties, use `.toMatchTypeOf`. This checks that an object
...
 ```
 
+`.toEqualTypeOf` and `.toMatchTypeOf` both fail on missing properties:
+
+```typescript
+// @ts-expect-error
+expectTypeOf({a: 1}).toEqualTypeOf({a: 1, b: 1})
+// @ts-expect-error
+expectTypeOf({a: 1}).toMatchTypeOf({a: 1, b: 1})
+```
```
trevorade, line 97: To be clear, could this also be written as?

```typescript
expectTypeOf({a: 1}).not.toEqualTypeOf({a: 1, b: 1})
expectTypeOf({a: 1}).not.toMatchTypeOf({a: 1, b: 1})
```

src/index.ts
```diff
@@ -3,6 +3,7 @@ export type Or<Types extends boolean[]> = Types[number] extends false ? false :
 export type And<Types extends boolean[]> = Types[number] extends true ? true : false
 export type Eq<Left extends boolean, Right extends boolean> = Left extends true ? Right : Not<Right>
 export type Xor<Types extends [boolean, boolean]> = Not<Eq<Types[0], Types[1]>>
+export type Ternary<B extends boolean, ValueIfTrue, ValueIfFalse> = B extends true ? ValueIfTrue : ValueIfFalse
```
trevorade: Is this still needed?

test/errors.test.ts
```diff
@@ -13,55 +13,56 @@ const tsErrors = (code: string) => {
...
   `)
 })
 
 test('toMatchTypeOf<...>() error message', async () => {
   expect(tsErrors(`expectTypeOf({a: 1}).toMatchTypeOf<{a: string}>()`)).toMatchInlineSnapshot(`
-    "test/test.ts:3:22 - error TS2554: Expected 1 arguments, but got 0.
+    "test/test.ts:3:36 - error TS2344: Type '{ a: string; }' does not satisfy the constraint 'Mismatch<{ a: number; }>'.
+      Type '{ a: string; }' is not assignable to type '{ a: number; }'.
+        Types of property 'a' are incompatible.
+          Type 'string' is not assignable to type 'number'.
```
trevorade, line 47: This error message appears backwards to me.

`expectTypeOf<Actual>().toMatchTypeOf<Expected>`

So I'd expect it to be:

> `Type 'number' is not assignable to type 'string'`

etc.

src/index.ts
mmkal: Nope, left in accidentally, deleted.

README.md
mmkal: Yes - at this stage of the "readme", `.not` hasn't been introduced yet so I opted for `// @ts-expect-error` since it makes very clear what a failing assertion looks like.

Also, I added a brief explanation of the slightly unusual way that the readme is generated - these tests _are_ real tests, but they are also the documentation for the library: https://github.com/mmkal/expect-type#contributing

trevorade:
Here's something kind of weird...

With these changes in the `dx-generics` branch, both the following compile:

```ts
  expectTypeOf({a: 1}).toMatchTypeOf<{a: number; b?: string}>()
  expectTypeOf({a: 1}).not.toMatchTypeOf<{a: number; b?: string}>()
```

In `main`, only the former compiles.

mmkal:
> Here's something kind of weird...
> 
> With these changes in the `dx-generics` branch, both the following compile:
> 
> ```ts
>   expectTypeOf({a: 1}).toMatchTypeOf<{a: number; b?: string}>()
>   expectTypeOf({a: 1}).not.toMatchTypeOf<{a: number; b?: string}>()
> ```
> 
> In `main`, only the former compiles.

Good catch, thank you. I've converted this to a branch since there are a few things up in the air and it isn't ready to go in.

trevorade:
Spent some time talking about this with @shicks and @12wrigja and we came up with some good ideas for some more dx.

Succinctly, it's a change like this:

```ts
const error = Symbol('error')

type ExpectedToBeNull<T> = {[error]: T}
type ExpectedNotToBeNull<T> = {[error]: T}

type IsASuperTypeOf<T> = {[error]: T}
type IsNotASuperTypeOf<T> = {[error]: T}

export type ExpectTypeOf<Actual, B extends boolean> = {
  toBeNull:
     Extends<Actual, null> extends B ?
         () => true :
         B extends true ? ExpectedToBeNull<Actual> : ExpectedToNotBeNull<Actual>

  toMatchTypeOf: {
    <Expected extends (
      B extends true ?
        (Extends<Actual, Expected> extends true ? unknown : IsASuperTypeOf<Actual>) :
        (Extends<Actual, Expected> extends false ? unknown : IsNotASuperTypeOf<Actual>)
    )>(): true
    <Expected extends (
      B extends true ?
          (Extends<Actual, Expected> extends true ? unknown : IsASuperTypeOf<Actual>) :
          (Extends<Actual, Expected> extends false ? unknown : IsNotASuperTypeOf<Actual>)
     )>(
      expected: Expected,
    ): true
  }
```

For `toBeNull`, you get errors like:
```
test/usage.test.ts:97:19 - error TS2349: This expression is not callable.
  Type 'ExpectedToBeNull<number>' has no call signatures.

97   expectTypeOf(1).toBeNull()
                     ~~~~~~~~

test/usage.test.ts:96:26 - error TS2349: This expression is not callable.
  Type 'ExpectedNotToBeNull<null>' has no call signatures.

96   expectTypeOf(null).not.toBeNull()
                            ~~~~~~~~
```

For `toMatchTypeOf`, you get an error message like:
```
test/usage.test.ts:27:49 - error TS2344: Type '{ a: string | number; }' does not satisfy the constraint 'IsNotASuperTypeOf<{ a: number; }>'.
  Property '[error]' is missing in type '{ a: string | number; }' but required in type 'IsNotASuperTypeOf<{ a: number; }>'.

27   expectTypeOf<{a: number}>().not.toMatchTypeOf<{a: number|string}>()
                                                   ~~~~~~~~~~~~~~~~~~

  src/index.ts:108:30
    108 type IsNotASuperTypeOf<T> = {[error]: T}
                                     ~~~~~~~
    '[error]' is declared here.

```

This loses tsc telling you specifically what wasn't matching between expected and actual as having the previous comparison be backwards was a bit misleading but it _does_ print out the types so you're aware of what types you were dealing with. Unfortunately, the only useful bit is the very first line of the error. Everything else is noise.

papb:
Hello, I didn't really look well at the PR, but just wanted to mention that I've noticed that the message that appears when hovering on an IDE can change significantly between TS versions. In particular, the expansions of generic expressions. Some versions may show `Foo<Bar>` while others may show things like `{ resultOfExpandingFoo: Bar }`. Especially with complex types (with many layers of generics). Sometimes, the result of expanding the expression was very simple, but still TS would show things like `X extends infer Y ? NonNullable<DeepPartial<Y>> : Something<X>` instead of resolving into `{ a: 1 }`. Just a reminder that this may happen. Maybe you put lots of effort in giving users good messages on hovering, but then a new TS version comes (every month!) and it could break...

shicks:
Seems like the easy solution here is to just make sure any possible expansion is readable.  In particular,

```
type IsASupertypeOf<T> = {[error]: T};
```

is not particularly readable if it's expanded, but if the internal symbol were instead

```
type IsASupertypeOf<T> = {[isASuperTypeOf]: T};
```

then it shouldn't really matter whether it's expanded or not - I can't imagine any reasonable way TypeScript could format that badly.

mmkal:
> Spent some time talking about this with @shicks and @12wrigja and we came up with some good ideas for some more dx.
> 
> Succinctly, it's a change like this:
> 
> ```ts
> const error = Symbol('error')
> 
> type ExpectedToBeNull<T> = {[error]: T}
> type ExpectedNotToBeNull<T> = {[error]: T}
> 
> type IsASuperTypeOf<T> = {[error]: T}
> type IsNotASuperTypeOf<T> = {[error]: T}
> 
> export type ExpectTypeOf<Actual, B extends boolean> = {
>   toBeNull:
>      Extends<Actual, null> extends B ?
>          () => true :
>          B extends true ? ExpectedToBeNull<Actual> : ExpectedToNotBeNull<Actual>
> 
>   toMatchTypeOf: {
>     <Expected extends (
>       B extends true ?
>         (Extends<Actual, Expected> extends true ? unknown : IsASuperTypeOf<Actual>) :
>         (Extends<Actual, Expected> extends false ? unknown : IsNotASuperTypeOf<Actual>)
>     )>(): true
>     <Expected extends (
>       B extends true ?
>           (Extends<Actual, Expected> extends true ? unknown : IsASuperTypeOf<Actual>) :
>           (Extends<Actual, Expected> extends false ? unknown : IsNotASuperTypeOf<Actual>)
>      )>(
>       expected: Expected,
>     ): true
>   }
> ```
> 
> For `toBeNull`, you get errors like:
> 
> ```
> test/usage.test.ts:97:19 - error TS2349: This expression is not callable.
>   Type 'ExpectedToBeNull<number>' has no call signatures.
> 
> 97   expectTypeOf(1).toBeNull()
>                      ~~~~~~~~
> 
> test/usage.test.ts:96:26 - error TS2349: This expression is not callable.
>   Type 'ExpectedNotToBeNull<null>' has no call signatures.
> 
> 96   expectTypeOf(null).not.toBeNull()
>                             ~~~~~~~~
> ```
> 
> For `toMatchTypeOf`, you get an error message like:
> 
> ```
> test/usage.test.ts:119:54 - error TS2344: Type '{ a: string; }' does not satisfy the constraint 'IsASuperTypeOf<{ a: string | number; }>'.
>   Property '[error]' is missing in type '{ a: string; }' but required in type 'IsASuperTypeOf<{ a: string | number; }>'.
> 
> 119   expectTypeOf<{a: string | number}>().toMatchTypeOf<{a: string}>()
>                                                          ~~~~~~~~~~~
> 
>   src/index.ts:108:27
>     108 type IsASuperTypeOf<T> = {[error]: T}
>                                   ~~~~~~~
>     '[error]' is declared here.
> ```
> 
> and for `.not`:
> 
> ```
> test/usage.test.ts:27:49 - error TS2344: Type '{ a: string | number; }' does not satisfy the constraint 'IsNotASuperTypeOf<{ a: number; }>'.
>   Property '[error]' is missing in type '{ a: string | number; }' but required in type 'IsNotASuperTypeOf<{ a: number; }>'.
> 
> 27   expectTypeOf<{a: number}>().not.toMatchTypeOf<{a: number|string}>()
>                                                    ~~~~~~~~~~~~~~~~~~
> 
>   src/index.ts:108:30
>     108 type IsNotASuperTypeOf<T> = {[error]: T}
>                                      ~~~~~~~
>     '[error]' is declared here.
> ```
> 
> This loses tsc telling you specifically what wasn't matching between expected and actual as having the previous comparison be backwards was a bit misleading but it _does_ print out the types so you're aware of what types you were dealing with. Unfortunately, there's quite a bit of noise in these.
> 
> I recognize that this snippet doesn't handle `unknown` and `never` correctly like it did before so you'd probably want to adjust for that as well.

@trevorade @shicks @12wrigja I did something like this and I think it's a big improvement! I ended up doing lots of refactoring, including separating out positive from negative assertions. Would you like to take a look?

mmkal:
I'm merging this and will release as v0.17.0 soon.

PR #18, opened 2022-10-28
Update DeepBrand to improve type checking capabilities

trevorade:
Makes `DeepBrand` more comprehensive to cover more cases. See the updated `usage.test.ts` test cases to get an idea of what is supported now.

Most things in TypeScript can have properties attached to them. There are also some things that `DeepBrand` can only check for by just directly including `value: T` such as private properties on classes.

### DIFF
--- a/test/types.test.ts
+++ b/test/types.test.ts
@@ -189,3 +189,160 @@ test(`undefined isn't removed from unions`, () => {
   expectTypeOf<string | null | undefined>().toEqualTypeOf<string | null | undefined>()
   expectTypeOf<string | null | undefined>().toMatchTypeOf<string | null | undefined>()
 })
+
+// Note: Works because of DeepBrand `type: 'function'` properties `params` and `return`.
+test('Distinguish between functions whose return types differ by readonly prop', () => {
+  type A1 = {readonly x: number}
+  type B1 = {x: number}
+
+  function f(a1: A1): A1 {
+    return a1
+  }
+
+  function g(b1: B1): B1 {
+    return b1
+  }
+
+  expectTypeOf<typeof f>().not.toEqualTypeOf<typeof g>()
+})
+
+// Note: Works because of DeepBrand `type: 'object'` property `value`.
+test('Distinguish between classes with only private properties', () => {
+  class A {
+    // eslint-disable-next-line mmkal/@typescript-eslint/class-literal-property-style
+    private readonly a = 1
+  }
+
+  class B {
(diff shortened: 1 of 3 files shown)


test/types.test.ts
```diff
@@ -189,3 +189,160 @@ test(`undefined isn't removed from unions`, () => {
   expectTypeOf<string | null | undefined>().toEqualTypeOf<string | null | undefined>()
   expectTypeOf<string | null | undefined>().toMatchTypeOf<string | null | undefined>()
 })
+
+// Note: Works because of DeepBrand `type: 'function'` properties `params` and `return`.
+test('Distinguish between functions whose return types differ by readonly prop', () => {
```
trevorade, line 194: This test case was passing before my changes to `index.ts`

All other test cases do not pass without the changes to `index.ts`

trevorade:
Note: I didn't author this in a branch within `expect-type` itself because, at present, I just have ssh pushing setup in my fork :)

mmkal (approved):
LGTM. Could probably refactor somewhat to get the keys-based props into a single place but not a big deal repeating a couple of times.

trevorade:
@mmkal 

I'd advise against releasing a branch at this time. I'm attempting to sync this to our codebase and I'm seeing some new errors like `Type instantiation is excessively deep and possibly infinite` which appear to be related to adding `readonly: ReadonlyKeys<T>` for `type: 'array'` in the `DeepBrand`.

I'll add another PR with a fix once I figure it out.

mmkal:
Thanks. Sounds like the type might be getting too big for typescript to handle. We should probably revert and add a test which reproduces this error.

trevorade:
Final investigation complete.

The issue was that the particular test case for `MergeDeep` had enough array nesting to trigger the TS compilers depth error when used in conjunction with `ReadonlyEquivalent` (though only the LHS, not the RHS oddly).

While debugging things closer, @shicks realized that `ReadonlyEquivalent` was in itself a complete replacement for `DeepBrand`. It ensures that the two types given to it are exactly equivalent eliminating the fairly expensive `DeepBrand`. We verified it on the `type-fest` + `expect-type` tests and it compiled much faster and everything worked as expected.

 I'll create a PR with that change for consideration.

mmkal:
Wow, great to hear! There is a commented out assertion in types.test.ts that might also be fixed by a more-efficient replacement for DeepBrand - if you open a PR I can uncomment and see if it works.

PR #20, opened 2022-11-03
Prevent the .not modifier from being chained

trevorade:
Chaining `.not` more than once is non-sensical so there is little benefit to support chaining it.

It is suppressed via the built-in `Omit`.


trevorade:
@mmkal, do you want to move forward with this change?

mmkal (approved):

PR #21, opened 2022-11-03
Rewrite `Equal` to use the equality check from `ReadonlyEquivalent` exclusively

### LINKED ISSUE #5: [expect-type] Bug with very deep type equality checking
[Playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBDAnmApnA3nAogRwK4CGANnAL5wBmUEIcA5CgB6oDGMAtEqnQNwBQfLmgAqARjgBebPmIAePnEVwAFAEpJAPgxwCAfgBcK9RK2YChgNqYARpYDOMKMAB2AcwA02loYdO3ZAF1AsjJ3BSU1TW09Q0iTaMsbe0cXDy8fFLceOAATAzhnPBBrFChg0iDyUj4NfiA)

```ts
import type { Equal } from 'expect-type';

type T1 = Equal<
    () => { a?: () => { a: [{ b: [string, { c: string }] }] } },
    () => { a?: () => { a: [{ b: [string, { c: string; d?: number }] }] } }
>;
//=> true
```

### LINKED ISSUE #26: Augmented functions aren't checked properly
This should pass:

```ts
expectTypeOf<(() => 1) & {x: 1}>().not.toEqualTypeOf<() => 1>()
```

Or more importantly this should fail:

```ts
expectTypeOf<(() => 1) & {x: 1}>().toEqualTypeOf<() => 1>()
```

### LINKED ISSUE #29: [bug] Type of nested properties are not fully resolved and fail equality test.
```ts
{
  type ObjectWithProp<PROP> = {
    prop: PROP;
  };
  type MakeNestObject<NEST_OBJECT extends { nestProp: ObjectWithProp<any> }> = {
    nestObj: NEST_OBJECT;
  };

  /* as expected */
  expectTypeOf<
    MakeNestObject<{ nestProp: ObjectWithProp<"one"> }>
  >().toEqualTypeOf<{ nestObj: { nestProp: { prop: "one" } } }>();
  expectTypeOf<
    MakeNestObject<{ nestProp: ObjectWithProp<"two"> }>
  >().toEqualTypeOf<{ nestObj: { nestProp: { prop: "two" } } }>();

  /* but */
  expectTypeOf<
    MakeNestObject<{ nestProp: ObjectWithProp<"one"> }>
  >().toEqualTypeOf<MakeNestObject<{ nestProp: ObjectWithProp<"two"> }>>();

  /* assignability test, the old fashion way */
  let o1: MakeNestObject<{ nestProp: ObjectWithProp<"one"> }> = null as any;
  let o2: MakeNestObject<{ nestProp: ObjectWithProp<"two"> }> = null as any;
  // @ts-expect-error - type "two" is not assignable to type "one"
  o1 = o2;
  // @ts-expect-error - type "one" is not assignable to type "two"
  o2 = o1;

  /* if the nested object type were by different name */
  type AnotherName<PROP> = {
    prop: PROP;
  }
  expectTypeOf<
    MakeNestObject<{ nestProp: ObjectWithProp<"one"> }>
  >().toEqualTypeOf<MakeNestObject<{ nestProp: AnotherName<"one"> }>>();
  expectTypeOf<
    MakeNestObject<{ nestProp: ObjectWithProp<"one"> }>
  >().not.toEqualTypeOf<MakeNestObject<{ nestProp: AnotherName<"two"> }>>();
}
```
[**Playground**](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgUwB5mQYxgFQJ4YDyAZnAL5zFQQhwBEaG2AtDAcnQNwBQ3C3cOGwxxCAIwBWWGAHVgMABYAFamAA8SgEqElAPjgBeRAMFwwqgFxwtOnoLJ2h7OAFkAhgGtkAOWQBnGHEpbDVvAFEAZRwAfUIAIQApMIBhHBRUGGQAOwATP0Q4LP8YFQgwKyDpOUVS9TcsvH0yfSN+U0LioKtwqNjElJxHB15BAHoAKjg3fMZpZBy4cdGTWex8ImI1E0F3L18AypCkIoDaiskq+WVVNToIIrom3RNdAAoASgA6GAgwgEcAK5uAA262QJDUx06kisUNOlgK5jKVjuD3I6OaH0cq1w7Ah21cnh80OCMEhHXhyNEF2w1WuZVuMAA7hBHuRnoI3l8fv8gaC8Zs4YEYQUTiUEUgkeV6MzWRj2ViRnAJnAxAD4EsVuhpGD8e1dsSDjSyUKztTSXTard7hwni8Pt9foCQbrNgb9sLSeSxWbDrIrla6LK2c0uY4TCrpn5gABzLJuMTAYHyPBCYoAGiECmQcAgwIWxGmCmA9zgTLcqc1gmByHgEAAjFZ3STpN7ir7jZabqjbezDIUAcDgVN8vU8I4a3WAExNokev1tynSv1dhlBlkhloDocjqYNRyjUZwAACMD8zBxF6g1CgcGYThE67lwHyWQg8CjsfjYhrQggD5zHs6BMBt+wgKcDyPU9z0vZBr2gO8APoG06DgF9CnfEdozjBNfx+JCn2AwRwLA+seAjSZgFIRQczFeZc2NJCmTgnMxFTHIqOIFisngeMQBzKskIAQTfGioG8Nx+I0bQ9H7NpTClKwbCUIYtSYXENi2fU5xbI4KXFKkVwDbsULtTkHR5Z1+U05sjS9U0ERE99s3EyTkGtB4njDNSdQFLTTFsz1WwcwzO2MtdTPZe0vlEx1eRdPzAoXELpScsSJKkwivMVMggA)

trevorade:
Fixes #29
Fixes #26
Fixes #5

> Note: I extracted a very small part of this PR to https://github.com/mmkal/expect-type/pull/20

This is a breaking change as I opted to remove the types that were no longer needed. They are exported though so it's likely some people depend on them. I can add these back as desired.

This took a lot of tinkering. This topic and this equality check is discussed extensively at https://github.com/Microsoft/TypeScript/issues/27024

The main three edge-cases this implementation worked around are:
1. Explicitly handling `any` separately
2. Supporting identity unions
3. Supporting identity intersections

The only remaining known issue with this implementation is:

```ts
  // @ts-expect-error This is the bug.
  expectTypeOf<{foo: number} & {bar: string}>().toEqualTypeOf<{foo: number; bar: string}>()
```

@shicks and I could not find a tweak to the `Equality` check to make this work.

Instead, I added a workaround in the shape of a new `.simplified` modifier that works similar to `.not`:

```ts
  // The workaround is the new optional .simplified modifier.
  expectTypeOf<{foo: number} & {bar: string}>().simplified.toEqualTypeOf<{foo: number; bar: string}>()
```

I'm not entirely sure what to do with documenting `.simplified` because it's something you should never use unless you need it. The simplify operation tends to lose information about the types being tested (e.g., functions become `{}` and classes lose their constructors). I'll definitely update this PR to reference the `.simplified` modifier but I wanted to get a review on this approach first. One option would be to keep around all the `DeepBrand` stuff and to have `.deepBranded` or something being the modifier instead. That would have the benefit of preserving all the exported types making this less of a breaking change.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -101,64 +101,68 @@ export type ConstructorParams<Actual> = Actual extends new (...args: infer P) =>
     : P
   : never
 
-type MismatchArgs<B extends boolean, C extends boolean> = Eq<B, C> extends true ? [] : [never]
+type MismatchArgs<ActualResult extends boolean, ExpectedResult extends boolean> = Eq<ActualResult, ExpectedResult> extends true ? [] : [never]
 
-export interface ExpectTypeOf<Actual, B extends boolean, Branded = false> {
-  toBeAny: (...MISMATCH: MismatchArgs<IsAny<Actual>, B>) => true
-  toBeUnknown: (...MISMATCH: MismatchArgs<IsUnknown<Actual>, B>) => true
-  toBeNever: (...MISMATCH: MismatchArgs<IsNever<Actual>, B>) => true
-  toBeFunction: (...MISMATCH: MismatchArgs<Extends<Actual, (...args: any[]) => any>, B>) => true
-  toBeObject: (...MISMATCH: MismatchArgs<Extends<Actual, object>, B>) => true
-  toBeArray: (...MISMATCH: MismatchArgs<Extends<Actual, any[]>, B>) => true
-  toBeNumber: (...MISMATCH: MismatchArgs<Extends<Actual, number>, B>) => true
-  toBeString: (...MISMATCH: MismatchArgs<Extends<Actual, string>, B>) => true
-  toBeBoolean: (...MISMATCH: MismatchArgs<Extends<Actual, boolean>, B>) => true
-  toBeVoid: (...MISMATCH: MismatchArgs<Extends<Actual, void>, B>) => true
-  toBeSymbol: (...MISMATCH: MismatchArgs<Extends<Actual, symbol>, B>) => true
-  toBeNull: (...MISMATCH: MismatchArgs<Extends<Actual, null>, B>) => true
-  toBeUndefined: (...MISMATCH: MismatchArgs<Extends<Actual, undefined>, B>) => true
-  toBeNullable: (...MISMATCH: MismatchArgs<Not<Equal<Actual, NonNullable<Actual>, Branded>>, B>) => true
+export interface ExpectTypeOfOptions {
+  positive: boolean
+  branded: boolean
+}
+export interface ExpectTypeOf<Actual, Options extends ExpectTypeOfOptions> {
+  toBeAny: (...MISMATCH: MismatchArgs<IsAny<Actual>, Options['positive']>) => true
+  toBeUnknown: (...MISMATCH: MismatchArgs<IsUnknown<Actual>, Options['positive']>) => true
+  toBeNever: (...MISMATCH: MismatchArgs<IsNever<Actual>, Options['positive']>) => true
+  toBeFunction: (...MISMATCH: MismatchArgs<Extends<Actual, (...args: any[]) => any>, Options['positive']>) => true
+  toBeObject: (...MISMATCH: MismatchArgs<Extends<Actual, object>, Options['positive']>) => true
+  toBeArray: (...MISMATCH: MismatchArgs<Extends<Actual, any[]>, Options['positive']>) => true
+  toBeNumber: (...MISMATCH: MismatchArgs<Extends<Actual, number>, Options['positive']>) => true
+  toBeString: (...MISMATCH: MismatchArgs<Extends<Actual, string>, Options['positive']>) => true
+  toBeBoolean: (...MISMATCH: MismatchArgs<Extends<Actual, boolean>, Options['positive']>) => true
+  toBeVoid: (...MISMATCH: MismatchArgs<Extends<Actual, void>, Options['positive']>) => true
+  toBeSymbol: (...MISMATCH: MismatchArgs<Extends<Actual, symbol>, Options['positive']>) => true
+  toBeNull: (...MISMATCH: MismatchArgs<Extends<Actual, null>, Options['positive']>) => true
+  toBeUndefined: (...MISMATCH: MismatchArgs<Extends<Actual, undefined>, Options['positive']>) => true
+  toBeNullable: (...MISMATCH: MismatchArgs<Not<Equal<Actual, NonNullable<Actual>, Options['branded']>>, Options['positive']>) => true
   toMatchTypeOf: {
-    <Expected>(...MISMATCH: MismatchArgs<Extends<Actual, Expected>, B>): true
-    <Expected>(expected: Expected, ...MISMATCH: MismatchArgs<Extends<Actual, Expected>, B>): true
+    <Expected>(...MISMATCH: MismatchArgs<Extends<Actual, Expected>, Options['positive']>): true
+    <Expected>(expected: Expected, ...MISMATCH: MismatchArgs<Extends<Actual, Expected>, Options['positive']>): true
   }
   toEqualTypeOf: {
-    <Expected>(...MISMATCH: MismatchArgs<Equal<Actual, Expected, Branded>, B>): true
-    <Expected>(expected: Expected, ...MISMATCH: MismatchArgs<Equal<Actual, Expected, Branded>, B>): true
+    <Expected>(...MISMATCH: MismatchArgs<Equal<Actual, Expected, Options['branded']>, Options['positive']>): true
+    <Expected>(expected: Expected, ...MISMATCH: MismatchArgs<Equal<Actual, Expected, Options['branded']>, Options['positive']>): true
   }
-  toBeCallableWith: B extends true ? (...args: Params<Actual>) => true : never
-  toBeConstructibleWith: B extends true ? (...args: ConstructorParams<Actual>) => true : never
+  toBeCallableWith: Options['positive'] extends true ? (...args: Params<Actual>) => true : never
+  toBeConstructibleWith: Options['positive'] extends true ? (...args: ConstructorParams<Actual>) => true : never
   toHaveProperty: <K extends string>(
     key: K,
-    ...MISMATCH: MismatchArgs<Extends<K, keyof Actual>, B>
-  ) => K extends keyof Actual ? ExpectTypeOf<Actual[K], B, Branded> : true
-  extract: <V>(v?: V) => ExpectTypeOf<Extract<Actual, V>, B, Branded>
-  exclude: <V>(v?: V) => ExpectTypeOf<Exclude<Actual, V>, B, Branded>
-  parameter: <K extends keyof Params<Actual>>(number: K) => ExpectTypeOf<Params<Actual>[K], B, Branded>
-  parameters: ExpectTypeOf<Params<Actual>, B, Branded>
-  constructorParameters: ExpectTypeOf<ConstructorParams<Actual>, B, Branded>
-  thisParameter: ExpectTypeOf<ThisParameterType<Actual>, B, Branded>
-  instance: Actual extends new (...args: any[]) => infer I ? ExpectTypeOf<I, B, Branded> : never
-  returns: Actual extends (...args: any[]) => infer R ? ExpectTypeOf<R, B, Branded> : never
-  resolves: Actual extends PromiseLike<infer R> ? ExpectTypeOf<R, B, Branded> : never
-  items: Actual extends ArrayLike<infer R> ? ExpectTypeOf<R, B, Branded> : never
-  guards: Actual extends (v: any, ...args: any[]) => v is infer T ? ExpectTypeOf<T, B, Branded> : never
+    ...MISMATCH: MismatchArgs<Extends<K, keyof Actual>, Options['positive']>
+  ) => K extends keyof Actual ? ExpectTypeOf<Actual[K], Options> : true
+  extract: <V>(v?: V) => ExpectTypeOf<Extract<Actual, V>, Options>
+  exclude: <V>(v?: V) => ExpectTypeOf<Exclude<Actual, V>, Options>
+  parameter: <K extends keyof Params<Actual>>(number: K) => ExpectTypeOf<Params<Actual>[K], Options>
+  parameters: ExpectTypeOf<Params<Actual>, Options>
+  constructorParameters: ExpectTypeOf<ConstructorParams<Actual>, Options>
+  thisParameter: ExpectTypeOf<ThisParameterType<Actual>, Options>
+  instance: Actual extends new (...args: any[]) => infer I ? ExpectTypeOf<I, Options> : never
+  returns: Actual extends (...args: any[]) => infer R ? ExpectTypeOf<R, Options> : never
+  resolves: Actual extends PromiseLike<infer R> ? ExpectTypeOf<R, Options> : never
+  items: Actual extends ArrayLike<infer R> ? ExpectTypeOf<R, Options> : never
+  guards: Actual extends (v: any, ...args: any[]) => v is infer T ? ExpectTypeOf<T, Options> : never
   asserts: Actual extends (v: any, ...args: any[]) => asserts v is infer T
     ? // Guard methods `(v: any) => asserts v is T` does not actually defines a return type. Thus, any function taking 1 argument matches the signature before.
       // In case the inferred assertion type `R` could not be determined (so, `unknown`), consider the function as a non-guard, and return a `never` type.
       // See https://github.com/microsoft/TypeScript/issues/34636
       unknown extends T
       ? never
-      : ExpectTypeOf<T, B, Branded>
+      : ExpectTypeOf<T, Options>
     : never
-  branded: Omit<ExpectTypeOf<Actual, B, true>, 'branded'>
-  not: Omit<ExpectTypeOf<Actual, Not<B>, Branded>, 'branded'>
+  branded: Omit<ExpectTypeOf<Actual, {positive: Options['positive']; branded: true}>, 'branded'>
+  not: Omit<ExpectTypeOf<Actual, {positive: Not<Options['positive']>; branded: Options['branded']}>, 'not'>
 }
 const fn: any = () => true
 
 export type _ExpectTypeOf = {
-  <Actual>(actual: Actual): ExpectTypeOf<Actual, true>
-  <Actual>(): ExpectTypeOf<Actual, true>
+  <Actual>(actual: Actual): ExpectTypeOf<Actual, {positive: true; branded: false}>
+  <Actual>(): ExpectTypeOf<Actual, {positive: true; branded: false}>
 }
 
 /**
@@ -183,7 +187,7 @@ export type _ExpectTypeOf = {
  * @description
  * See the [full docs](https://npmjs.com/package/expect-type#documentation) for lots more examples.
  */
-export const expectTypeOf: _ExpectTypeOf = <Actual>(_actual?: Actual): ExpectTypeOf<Actual, true> => {
+export const expectTypeOf: _ExpectTypeOf = <Actual>(_actual?: Actual): ExpectTypeOf<Actual, {positive: true; branded: false}> => {
   const nonFunctionProperties = [
     'parameters',
     'returns',
@@ -230,5 +234,5 @@ export const expectTypeOf: _ExpectTypeOf = <Actual>(_actual?: Actual): ExpectTyp
   const getterProperties: readonly Keys[] = nonFunctionProperties
   getterProperties.forEach((prop: Keys) => Object.defineProperty(obj, prop, {get: () => expectTypeOf({})}))
 
-  return obj as ExpectTypeOf<Actual, true>
+  return obj as ExpectTypeOf<Actual, {positive: true, branded: false}>
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -84,15 +84,17 @@ type ReadonlyEquivalent<X, Y> = Extends<
 export type Extends<L, R> = IsNever<L> extends true ? IsNever<R> : [L] extends [R] ? true : false
 export type StrictExtends<L, R> = Extends<DeepBrand<L>, DeepBrand<R>>
 
-type StrictEqual<L, R> =
-  (<T>() => T extends (L & T) | T ? true : false) extends
-  (<T>() => T extends (R & T) | T ? true : false) ?
-    IsNever<L> extends IsNever<R> ? true : false : false
+type StrictEqual<L, R> = (<T>() => T extends (L & T) | T ? true : false) extends <T>() => T extends (R & T) | T
+  ? true
+  : false
+  ? IsNever<L> extends IsNever<R>
+    ? true
+    : false
+  : false
 
-export type Equal<Left, Right, Branded = true> =
-  Branded extends true
-    ? And<[StrictExtends<Left, Right>, StrictExtends<Right, Left>]>
-    : StrictEqual<Left, Right>
+export type Equal<Left, Right, Branded = true> = Branded extends true
+  ? And<[StrictExtends<Left, Right>, StrictExtends<Right, Left>]>
+  : StrictEqual<Left, Right>
 
 export type Params<Actual> = Actual extends (...args: infer P) => any ? P : never
 export type ConstructorParams<Actual> = Actual extends new (...args: infer P) => any
@@ -101,7 +103,12 @@ export type ConstructorParams<Actual> = Actual extends new (...args: infer P) =>
     : P
   : never
 
-type MismatchArgs<ActualResult extends boolean, ExpectedResult extends boolean> = Eq<ActualResult, ExpectedResult> extends true ? [] : [never]
+type MismatchArgs<ActualResult extends boolean, ExpectedResult extends boolean> = Eq<
+  ActualResult,
+  ExpectedResult
+> extends true
+  ? []
+  : [never]
 
 export interface ExpectTypeOfOptions {
   positive: boolean
@@ -121,14 +128,19 @@ export interface ExpectTypeOf<Actual, Options extends ExpectTypeOfOptions> {
   toBeSymbol: (...MISMATCH: MismatchArgs<Extends<Actual, symbol>, Options['positive']>) => true
   toBeNull: (...MISMATCH: MismatchArgs<Extends<Actual, null>, Options['positive']>) => true
   toBeUndefined: (...MISMATCH: MismatchArgs<Extends<Actual, undefined>, Options['positive']>) => true
-  toBeNullable: (...MISMATCH: MismatchArgs<Not<Equal<Actual, NonNullable<Actual>, Options['branded']>>, Options['positive']>) => true
+  toBeNullable: (
+    ...MISMATCH: MismatchArgs<Not<Equal<Actual, NonNullable<Actual>, Options['branded']>>, Options['positive']>
+  ) => true
   toMatchTypeOf: {
     <Expected>(...MISMATCH: MismatchArgs<Extends<Actual, Expected>, Options['positive']>): true
     <Expected>(expected: Expected, ...MISMATCH: MismatchArgs<Extends<Actual, Expected>, Options['positive']>): true
   }
   toEqualTypeOf: {
     <Expected>(...MISMATCH: MismatchArgs<Equal<Actual, Expected, Options['branded']>, Options['positive']>): true
-    <Expected>(expected: Expected, ...MISMATCH: MismatchArgs<Equal<Actual, Expected, Options['branded']>, Options['positive']>): true
+    <Expected>(
+      expected: Expected,
+      ...MISMATCH: MismatchArgs<Equal<Actual, Expected, Options['branded']>, Options['positive']>
+    ): true
   }
   toBeCallableWith: Options['positive'] extends true ? (...args: Params<Actual>) => true : never
   toBeConstructibleWith: Options['positive'] extends true ? (...args: ConstructorParams<Actual>) => true : never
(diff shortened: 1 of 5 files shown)


trevorade:
Just an anecdote, this PR is compatible with all of the `type-fest` build tests (in our repo, we have an automation step which rewrites the `tsd` expectations to use `expectTypeOf`).

My previous reverted PR ran afoul of one of `merge-deep.ts` test cases which had too much nesting. This PR does not have that problem.

papb:
Hi @trevorade! I am against `.simplified`, for a few reasons (pick the one that convinces you the most, I guess).

1. I like to consider `{ a: 1 } & { b: 2 }` identical to `{ a: 1, b: 2 }`. Not just "simplifiedly"-identical. Really identical. I have not seen to this day any observable difference between the two. Therefore, I think `expectTypeOf<{ a: 1 } & { b: 2 }>().toEqualTypeOf({ a: 1, b: 2 })` must work, if not, it is a bug.

2. With this PR, given a type `T`, there is not a single correct way of testing the equality between the types `{ a: 1 } & { b: () => true }` and `{ a: 1, b: () => T }`. If `T` is true, `.simplified` must be used. If `T` is anything else, `.simplified` must not be used.

3. I may agree with what you said that "_you should never use unless you need it_", but the only way for someone to "know" they need it is if they've previously manually analyzed the types, in which case the test becomes meaningless at best (since it is just "masking" an actual manual test someone did - and can get outdated).

4. From what I've seen from the source code of your PR, it seems your implementation of `Simplified<T>` is not recursive, so it wouldn't catch `{ a: { x: 1 } & { y: 2 } }` and `{ a: { x: 1, y: 2 } }`. Even though this example can be fixed easily, I don't see any easy way to cover all complex recursive cases, such as intersections in return types of functions, high-order functions, the special thisArg, and so on.

5. With this PR, `expectTypeOf<A>().simplified.toEqualTypeOf<B>()` would pass for `type A = () => true` and `type B = Record<never, number>`. To me this is unacceptable, and makes `.simplified` impossible to be trusted, therefore I would never write a test using it.

papb:
> 1. I like to consider `{ a: 1 } & { b: 2 }` identical to `{ a: 1, b: 2 }`. Not just "simplifiedly"-identical. Really identical. I have not seen to this day any observable difference between the two. Therefore, I think `expectTypeOf<{ a: 1 } & { b: 2 }>().toEqualTypeOf({ a: 1, b: 2 })` must work, if not, it is a bug.

By the way, I would think we had a test for this already, but it seems we don't (since I didn't see you commenting out this test on the PR changes)?

papb:
Also, maybe I should have asked this before anything else, but why did you want to make this change? Is it just to work around #5?

shicks:
@papb I agree that `Simplify` as written definitely has some trustability problems.  But I'd also point out that `DeepBrand` is not without issues as well.  We've patched a number of them, but the fact remains that it's a very kludgy hodgepodge that's both very inefficient (we found massive speed-ups with using the new direct comparison version) and yet still has potential holes (i.e. readonly properties in parameters/returns of overloaded functions simply can't be handled correctly).

I could see making the `DeepBrand`-based version available when needed (instead of just `simplified`).  It unfortunately means we can't delete this very complicated bit of code that's only used for the one thing - but it would provide an alternative when the direct comparison has problems.  I do strongly believe the direct equality check is the best default, since it's basically just deferring the check directly to the type checker, rather than hacking something together.

mmkal:
Sorry for the delay in getting to this, hoping to find time to dive in in the next week. 

@shicks when you say direct equality check what do you mean exactly? TS doesn't have that built in as far as I'm aware, all we really have to play with is `extends`. Could you give an example of how you'd want it to look?

Definitely agreed with the general concern with how DeepBrand has grown and grown, slowly taking over everything. And every edge case seems to add to its complexity, making it harder to maintain and most likely slowing the compiler down. It started as a way to distinguish between hairy types like any/unknown/never - which I do think is really important and should be the default behaviour. This library wouldn't be doing its job if it let anys make things go green secretly.

trevorade:
I need to find some time to update this draft. I think what I'd want to do would be to keep around `DeepBrand` but have it not be the default. We could also un-revert my previous PR which improves `DeepBrand` (the problems there were not real problems, just an excessive amount of type nesting in the particular test I was looking at). Retaining `DeepBrand` also has the benefit of keeping the API of exported types stable.

Accessing the `DeepBrand` variant could be done with something like:

```ts
expectTypeOf<Foo>().branded.toEqualTypeOf<Bar>();
```

Honestly, nearly all examples we could find worked fine with this PR's new equality check with the exception of the intersection case noted above. This PR is more performant and covers several cases `DeepBrand` does not. So I think it's still a win.

TS really just needs an equality check in their typing system :stuck_out_tongue:

trevorade:
Updated to keep around `DeepBrand`. Removed `Simplified`. Added a `.branded` chain.

This has the same ergonomic issues as the `.simplified` chain.

Main benefit here is that all the other exported types beyond `expectTypeOf` are now unchanged with this commit. `expectTypeOf` _is_ a breaking change though in the off chance that people are checking one of the few cases where this new version fails.

As an alternative to the `.branded` chain, how would people feel about adding `expectBrandedTypeOf` as an exported function?

trevorade:
@mmkal, @papb Please take another look.

mmkal:
I really like this. I still need to go through the implementation, but I have very high confidence based on the tests, and not being (properly) breaking makes it much easier to say "yes" to. This is definitely a net-good change.

A couple of comments/requests:

- could you add a simplified version of the `.branded` test to usage.test.ts? maybe with inline comments so they appear on the readme, including a link to the typescript issue? Or could even put a markdown link as the rest title and it'll render nicely on the readme. I like the pattern of treating usage.test.ts as features that users should know about, and types.test.ts being edge-cases that should "just work". We could still keep the detailed test you have in types.test.ts if there's a simplified version that'd be more presentable (e.g. no need to test `expectTypeOf<...>()` and `expectTypeOf(...)` in usage.test.ts)
- I'm fine with this kind of very small breaking change. The lib is v0 anyway so not even a semver violation to publish as a new minor.
- could you update the PR description so it reflects what this change now does? I've set the squashed-commit to use the PR body as git commit message so will be a useful breadcrumb for people who update and get hit by new failures.

src/index.ts
```diff
@@ -84,7 +84,15 @@ type ReadonlyEquivalent<X, Y> = Extends<
 export type Extends<L, R> = IsNever<L> extends true ? IsNever<R> : [L] extends [R] ? true : false
 export type StrictExtends<L, R> = Extends<DeepBrand<L>, DeepBrand<R>>
 
-export type Equal<Left, Right> = And<[StrictExtends<Left, Right>, StrictExtends<Right, Left>]>
+type StrictEqual<L, R> =
+  (<T>() => T extends (L & T) | T ? true : false) extends
```
mmkal, line 88: Is there a link to a stack overflow/blog post/docs page that explains this trick or something similar? And/or an explanation of why we can't just use `T`. It's befuddling(-looking) enough to warrant a comment.

src/index.ts
```diff
@@ -95,7 +103,7 @@ export type ConstructorParams<Actual> = Actual extends new (...args: infer P) =>
 
 type MismatchArgs<B extends boolean, C extends boolean> = Eq<B, C> extends true ? [] : [never]
 
-export interface ExpectTypeOf<Actual, B extends boolean> {
+export interface ExpectTypeOf<Actual, B extends boolean, Branded = false> {
```
mmkal, line 106: Maybe we should merge in a separate PR that renames `B` to `Positive` or something? I feel like I've seen that elsewhere on one of the PRs recently, am I remembering that right?

mmkal (commented):
Some minor comments on the implementation part

papb:
> yet still has potential holes (i.e. readonly properties in parameters/returns of overloaded functions simply can't be handled correctly)

I didn't know this, can someone show an example please?

papb:
> 1. I like to consider `{ a: 1 } & { b: 2 }` identical to `{ a: 1, b: 2 }`. Not just "simplifiedly"-identical. Really identical. I have not seen to this day any observable difference between the two. Therefore, I think `expectTypeOf<{ a: 1 } & { b: 2 }>().toEqualTypeOf({ a: 1, b: 2 })` must work, if not, it is a bug.

I still insist on this... Can someone try to convince me otherwise?

papb:
> I have very high confidence based on the tests, and not being (properly) breaking makes it much easier to say "yes" to.

@mmkal The reason I don't agree with this sentence is that the test for `expectTypeOf<{ a: 1 } & { b: 2 }>().toEqualTypeOf({ a: 1, b: 2 })` should exist in the first place. Since it doesn't, which I think is a mistake, this PR gives the impression of not being a breaking change.

test/types.test.ts
```diff
@@ -174,8 +174,7 @@ test('parity with IsExact from conditional-type-checks', () => {
 test('Equal works with functions', () => {
   expectTypeOf<a.Equal<() => void, () => string>>().toEqualTypeOf<false>()
   expectTypeOf<a.Equal<() => void, (s: string) => void>>().toEqualTypeOf<false>()
-  // todo: workaround https://github.com/microsoft/TypeScript/issues/50670 - https://github.com/mmkal/expect-type/issues/5
-  // expectTypeOf<a.Equal<() => () => () => void, () => () => () => string>>().toEqualTypeOf<false>()
+  expectTypeOf<a.Equal<() => () => () => void, () => () => () => string>>().toEqualTypeOf<false>()
```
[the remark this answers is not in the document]
trevorade: Outdated though...

If you did the `.Equal` not branded, it _would_ work

trevorade:
@papb Note that all of the tests I add in this PR fail before this change. These tests also passed with my previous PR 304a54bbd8531bbeb28260565835b0bd84e4a492 which improved `DeepBrand` but the complexity of `DeepBrand` started hitting the ceiling of how much work `tsc` can do in type-land. I reverted it in a6d1f99797f7c0be48a84bc5258f61ee2472f5cd because it was causing tests to fail for `type-fest` with some of their types (we use `expect-type` instead of `tsd` to run the `type-fest` tests in our codebase).

`DeepBrand` is just very unwieldily and slow.

We're in a situation with: pick 2:

1. Fast
2. Covers most cases
3. Covers `{a: 1} & {b: 2}`  equals `{a: 1, b: 2}`

This PR gets you 1. and 2.

We could un-revert 304a54bbd8531bbeb28260565835b0bd84e4a492 and we'd presumably get 2. and 3. but at the cost of compile speed and not working with more complex types.

----

Lastly, note that even my previous PR doesn't work around https://github.com/microsoft/TypeScript/issues/50670 whereas this PR does.

It's pretty crappy that TS just doesn't have an equality check for types. At the moment, there is just no perfect equality check that covers _all_ cases. I don't really agree that 3. is table stakes though.

src/index.ts
trevorade: I'm guessing you're thinking of https://github.com/mmkal/expect-type/pull/13

In that one, I renamed it to `ExpectTypeOf<Actual, ExpectedResult extends boolean>`

I could incorporate parts of that PR to just the generic renaming.

shicks:
@papb wrote:
> > yet still has potential holes (i.e. readonly properties in parameters/returns of overloaded functions simply can't be handled correctly)
> 
> I didn't know this, can someone show an example please?

```
{
  // Overloads indistinguishable from non-overloads.
  interface A {
    foo(a: string): string;
    foo(a: number): number;
  }
  interface B {
    foo(a: number): number;
  }
  expectTypeOf<A>().toEqualTypeOf<B>(); // should fail, but passes
}
{
  // Generics indistinguishable from non-generics.
  interface A {
    foo<T>(a: T): T;
  }
  interface B {
    foo(a: unknown): unknown;
  }
  expectTypeOf<A>().toEqualTypeOf<B>(); // should fail, but passes
}
{
  // Readonly properties in parameters indistinguishable.
  interface A {
    foo(a: {readonly a: number}): number;
  }
  interface B {
    foo(a: {a: number}): number;
  }
  expectTypeOf<A>().toEqualTypeOf<B>(); // should fail, but passes
}
{
  // These are actually the same type, but DeepBrand can't figure it out, either.
  type Digit = 0|1|2|3|4|5|6|7|8|9;
  type A = `${number}${string}`;
  type B = `${Digit}${string}`;
  expectTypeOf<A>().toEqualTypeOfBranded<B>(); // should pass, but fails
}
```
Here are some examples of how the current `DeepBrand`-based implementation doesn't work as you might hope/expect ([playground link](https://www.typescriptlang.org/play?jsx=0&ssl=161&ssc=1&pln=1&pc=1#code/KYDwDg9gTgLgBDAnmYcByEYB4AqdQzAB2AJgM5wBGEEANsAIZEB8cAvHHgceQlAK6oA-HABmDWmVQAuPoIBQoSLATJUAeSi41FbqQrU6jIgG0Auqw44dJovwC2lYFDP4QhfWIlS4I8ZJk5YEVwaHgkFDgAQVJtFF13HgMaeiZzS04bO0dnVz1eGAFhILhZfykQ5XC1OABRAEcsABlgUXh85KMmABo4ACUASwBzAAt2xM9DVJZ2OBa2tw8Cot9+4bHS9EwsQdGYZkqw1UiADWg44ASlihMp4167pgtZjGwGi7ITAAYzXut4kwARgsB3kJGAAGNaAwoKgIRAiGR4FIIbCYLJ+EQBvVBHAyIhHHR5BFUABlSFo2YkiCiPEU4AweSHFQkuAASTIaGAADdnLgMiYcHkJrxbDzcqtCriyt5gkojqyOTFEPzZoLhdc4CZyaiGa4RK8sByubytDhmKwZQFmdVIhyAKpEADWRAgAHciKqOCZMS73UQNUktULVoalUQVebLV5rfKWTVjeKoJplV64JosCZE6b+b1w5HmBYmQB6ABUpfkcFL-Uh-CgZAGvNoiDgbokTrgDGOqCYJDgsLA0IhqAG8Ddo5GnaoUF7wD7rNh0MI84gCBGqGgwwGRAkADpOCMBhQj3B+FJRPxaGJoJXq8AcRJR4gALQQ9cQp3boZ4woDCGEKA1yYOAAAMokWIMACFVmgjpoklFYrR8JDgBAh5IQYM8R3gcFCH-MhbzXVASAGURRGcYhhyoBk3WAYgp3It18BIIZgFfBgfBJCgYBGBh4G5JgBloaFuzIVEBjAHCIEuIgAHJ4EoFtwXES90UI59QKYRAQLgbkKBA31XQ9ED1NAgBvftGBIBFm07WQkSgL84AAXx0vTzLsn9HKIb9XNMkCLIYIR7N-HyXLc-TApC7zvwAH1PUhWm3OdwtvYsbW7OAABE6LASCZ1ic1ZmzPkirgqVgjgVYzJJWRZKIJNZOcytNnzVVyqKFqRBqtQ6q0pqWtkB1nSMz0ypFbjOqq7rargWTDP9AaqtkLgJq8pz4uyJxAPix4iDgeL8UJK9dq3Ih4E2y8ToS5Tkr7eLuQgAYSC6uAzJaqqqtm2SwEc+xR0bYBZI+z6BNoQQVpB5rls4CDPAapiAApdxRmEhjIWQtPMABKdhWC0173s+z7vvhREpX-aBgeJqqwBhBh7AxuAAGEEQc-hKagAAFenGf5EGqu3JEmGHWQcuAPKCpII1yZF4B-mALBancGd-1wXoEbgZHUagdHMYjXG2HxiMLQOYnoc2VbNW13c0aZ7dyMArnDdYB3nH6OBi2LTtHuesRMX-AY2dbCc4BIsiKPOuA6ZnRnizROt9rBwQKHsPi3y-QmBe7OqLyIQOEWpmmY4Zpnxcl3ssC5s2aYTqAiDF3L8srvoa+Jnij0biXm8Kw8yB52OGWcBX+TblzBthuCsbMLOaZzuaYRnRAi+J0dgEZ2QzJMABpOBtzgJ1gEQGlODMFad7MC2qotzfs++iBKAAK0hGAV8+36IBQWABkuTed73-ah9j60iFF3CuhUL7MCvp9WEDBrJEGbLIPoVkbKIG3kfMg-M56whxAMWEJAkH3n4Hguc6DECYPNNnT+MAg47loLIdQklaESDIRQseVUybs05gPUuYCe7S1ZuTAQ3DeZsLHs1DKrJkG4PwawtMytCgMDVi1Imn1-77yASfIUz5gpvWcnDXgXM-xOnVnAberARANVNJsbeLVnImE0SA34LVHGcHkAcOMtoNBMIRCwjB8iQBQn4OCLAricC9GkcQ2R-iowlm9nAhgkkUqiCgBAeweI0k9kRLResCBVxjBgGADGXshgTn4JQXc8J7DFhJK+XiwliCsTIDUtQdSJD0B8pcYsR4yApyEPUNg0InC0AAKQAGYogAFYADUQzgCjImUwMgOT5Be2jmiGhzhnzDFdLCSRNRkFwNQXIoqHAFGq2wKo9RgCj5aLMDowhRyEGIAaMQsGxBsAgy3gAfV3vvbeZ9OAX2ct0L5z5YHwNsiYX5ACzGApwMClqrAOq4ksUmGxABueQ9iwm-APrckBzAsWrO9r9BkmyoDbKGLs4IUiUHPNeY2CQHysAnF6AATQyAopIWAWqI35IjF2k81onAQtKGMUhsagqqvy80gq8bCs1OysVgRyjAGxu4pknjMrcv0M0CJGQSpaCaMitaFVVhGp2NGEwTRAyeBMH0fUJQUL7MiKSX8-5dXkH1f0LlE0sDl34c0ZgvRA1SytaCbVrJ+FzleRIZorQYARPWPsWYMRpbag9TAL1mD5hJrWHsENcB3WOU9f63YYxeh5sLB40I8ZIjKyUdmh8tAE1tGTYW2YOasACqFVbIMebViAk2F8XovaFX9s8BW+AIhh2yC+ESrVdavFwB4XzKI-5+ASAyBumAW6rxwRtnbWQbsnZCq0qsLmmwrHOFdagQRXCYDQDXZg3d+6d2bokAYigmsj263tkQR2q7z0Rlem+r9cFf0gcQCDEQV74rmBBrILmE8b1QCZKyAAskeNOMA3xRH-VgWCa09q9GZt+qgKRjBcsaJBMjprNTmpEOYTYYpTQzwytuAC4gqLKxQP+BW6hRBYHA7QXoxHNR7VYKop9cbaCCdEDGghb0QZK3AC-OczAbaYbZKSTDUQcDMwABKyGw2QXD+HCNKbkyJz9Ym6jqf-Jp8TzBsayAqqpvjGmSBaaUN52QXmnMkF6CjXcOm9MGeM6ZnD6cRgEfRkRqWsaW22b3RIXogXlxFsgq59zU1x5fQgHJhTjb-y32Jmp-jWXtO6f04ZkzcAzMWbi4R0rzb92pf3RlxzWWXNuaCJ5nrmnEZ+aCwFobwW4ChfC3VqLjWYt4ZawltrNnRPdaq85uAOX+seevtipdVRMrfMywp2YqjOvbsRk2-dshRP9ZO2oITF37MVRrs9rT92esKee70V7+3wRQhhHCNm4wqsKdkMdr7j3RDEtUWs9QppaAQDgceUgR4aE+WIWQXilB6BiFSek10RBnwQER8j8gu4WpcecDx1A4FVFVVEDQK70Uvz9Ycl+LFxMmcQBZ3ALazh+sC6gFzgrADuNKNQNBBn15ecMFkMLoXDhtqi4tqNmA32ogfd3LJltCmlNEY+6L9Xmvte6-3SVkATbDeCqxRIuH3sADixBnB-lRyRJEX4sc47xyktJ-OETPlYg1UtZBKeC3OjTyX8EZc84FfLzg-WcCq6p5HqAtOtsqe58zhPC0PT9bz0QFPVUTfQ5E2borevocG5y7blqpeUBPa14KnXleLfQ7azb7GdumQO5rE82yH8v6bNR9HXmQ9cnbg9xjoY3uGC4+AOH8XUeqL05BjzvnZkIWoM8sL5ySucgi7saniXVFpfr5z5vBPe+D8q+PyXqHjfhPN+xq34r1ekvS1r93+vj-gBN4r3fyf072-x73kD7xwHXB8CB07Ds1sh4lQDIAZlQBJAeH4HgDDV7DgAhCYHkjEGGDrGwjgAgHQN6GAAnGcCX1ZCyi3HgA4C+FikBFigACZYoxlYoAAWWKSZWKAANligAHZYoAAOWKAATlF1ZHAg4BAgABIzI995COcfJXJJCahoIZD5CaDSkYBnIlDQohhVDf8wcy8X838q8n8a8jdjCX5TcW9zcJBLdrdQD9sgA)).

> > I have very high confidence based on the tests, and not being (properly) breaking makes it much easier to say "yes" to.
> 
> @mmkal The reason I don't agree with this sentence is that the test for `expectTypeOf<{ a: 1 } & { b: 2 }>().toEqualTypeOf({ a: 1, b: 2 })` should exist in the first place. Since it doesn't, which I think is a mistake, this PR gives the impression of not being a breaking change.

Ultimately this comes down to "what does _equal_ mean?"  It's true that I'm not aware of any _other_ way to distinguish `{a:1}&{b:2}` from `{a:1,b:2}`, aside from the one here.  But clearly TypeScript represents them differently internally, or else we wouldn't have this problem.  In most cases, TypeScript treats optional/readonly as irrelevant, so there's a slippery slope of which reductions you want to allow in "equals" versus which you want to preserve.  As we're all aware, there is no good way to ask the type checker if these two types will be treated exactly identically.  It's probably not even a decidable question, given that the type system is apparently Turing complete.  So I think it makes a lot of sense to expose the lowest-level check we possibly can, and then if you want to force it to apply certain transformations (i.e. look at it on a per-field level, etc) then you can opt into that explicitly.

trevorade:
@mmkal Another proposal.

We remove `.branded` and `DeepBrand` and add the following:

```ts
export type MutuallyAssignable<Actual, Expected> =
    [Actual, Expected] extends [Expected, Actual] ? true : false;
...
export interface ExpectTypeOf<Actual, B extends boolean> {
...
  toAssignMutuallyWithTypeOf: {
    <Expected>(...MISMATCH: MismatchArgs<MutuallyAssignable<Actual, Expected>, B>): true
    <Expected>(expected: Expected, ...MISMATCH: MismatchArgs<MutuallyAssignable<Actual, Expected>, B>): true
  }
```

This doesn't have protections against `any` or `never`  because it doesn't need to. `let a: any` and `let b: number` are mutually assignable to each other.

In the off-case that you run into something like `{a: 1} & {b: 2}` not being equal to `{a:1, b:2}`, you can use `toAssignMutuallyWithTypeOf`. It's a similar check but not as strong as `toEqualTypeOf`.

This simplifies `expect-type` quite a bit by removing `DeepBrand` and removes the confusion of the `.branded` chain method. If we do not an alternative to `toEqualTypeOf`, I think it would be best were it as simple and non-magic as possible. I think `MutuallyAssignable` fits that bill.

mmkal:
> > I have very high confidence based on the tests, and not being (properly) breaking makes it much easier to say "yes" to.
> 
> @mmkal The reason I don't agree with this sentence is that the test for `expectTypeOf<{ a: 1 } & { b: 2 }>().toEqualTypeOf({ a: 1, b: 2 })` should exist in the first place. Since it doesn't, which I think is a mistake, this PR gives the impression of not being a breaking change.

@papb good point, and sorry to have forgotten this. I'm currently on paternity leave so git history will be much more reliable than my memory. 

@trevorade in the tradeoff between fast and correct I'd prefer correct, but it does seem that "slow" also exposes tsc limitations which also means "not correct".

To make clear the tradeoff, @papb would you be open to a simple PR adding a test for the `{a: 1} & {b: 2}` case? That would force this PR to at least disable that test, explicitly, so we could see what we're losing.

A couple more thoughts that come to mind:

1. We could support both modes, and use some kind of pseudo-higher-kinded type via module augmentation to allow advanced users to opt in to a speed boost if they are prepared to sacrifice `&` working properly. I did something similar here: https://github.com/mmkal/handy-redis/blob/39f4f823c7d169491f93aa087fca7d122fe9d016/src/node_redis/multi.ts#L7-L22 and fp-ts has a more advanced implementation.
2. In either case we should probably have DeepBrand or its replacement force an error when we know it's going "too deep" for tsc to handle so we don't get bad types being marked as ok.

For either one, we should aim to have it "just work" for all the simple use cases in usage.test.ts if we possibly can. We might also want a limitations section of the documentation (maybe also generated from tests) so we can be honest and clear about what problems you might hit once you start trying to test big, complex types.

mmkal:
@trevorade I finally had a chance to dig into this (and experiment with an alternative in #27). Everything I was able to come up with that _technically_ covers all edge cases had serious performance problems, so because this implementation is so dramatically much simpler, I'm inclined to agree that we just have the `&` problem as a documented limitation. I also think that in most cases there are simpler workarounds than `branded` though. For example this workaround is available in simple cases:

```ts
type A = {a: 1}
type B = {b: 1}
type Intersection = A & B
expectTypeOf<Intersection>().toEqualTypeOf<{a: 1; b: 1}>()
expectTypeOf<{[K in keyof Intersection]: Intersection[K]}>().toEqualTypeOf<{a: 1; b: 1}>()
```

Is that what the original `simplified` was doing, more or less? I can't find the original commit.

trevorade:
It's been a while so I can't remember what `.simplified` was doing exactly. It's annoying that GitHub does not preserve older PR versions.

This current state of this PR has `.branded`. I think it would be better to clean up this file as much as possible and just remove the brand-related code.

Unfortunately, I don't really have capacity these days to help out with changes here. Feel free to create a new commit based on this or take over this one (if you can?)

mmkal:
Sounds good - will see if I can simplify, and merge when it's ready

gomain:
When will this be released?

unional:
> Covers {a: 1} & {b: 2} equals {a: 1, b: 2}

Just for sharing, I have it working in [type-plus](https://github.com/unional/type-plus/blob/main/ts/equal/equal.is_equal.spec.ts#L348)

mmkal:
> > Covers {a: 1} & {b: 2} equals {a: 1, b: 2}
> 
> Just for sharing, I have it working in [type-plus](https://github.com/unional/type-plus/blob/main/ts/equal/equal.is_equal.spec.ts#L348)

@unional while that does cover the specific test that was added to this PR to document the limitation, I don't think I would like a workaround like that to be included in expect-type. Really, it's just kicking the can down the road where it's less visible and harder to debug, since it's just explicitly looking for top-level intersections and handling them manually. Nested properties with intersection types will meet the same problem. That may be ok for your use case - I'm not familiar with type-pluts - but I think it'd be confusing here. I opened a PR to demonstrate/document this: https://github.com/unional/type-plus/pull/302

mmkal:
@gomain / others on this thread - this was released in [v16](https://github.com/mmkal/expect-type/releases/tag/v16.0.0)

PR #51, opened 2024-02-09
Add `.pick` and `.omit`

aryaemami59:
This PR:

  - [X] Adds `.pick` and `.omit` methods which work similarly to `.extract` and `.exclude`.


test/usage.test.ts
```diff
@@ -401,3 +401,15 @@ test('Another limitation: passing `this` references to `expectTypeOf` results in
...
+  type Person = {name: string; age: number}
+
+  expectTypeOf<Person>().pick<'name'>().toEqualTypeOf<{name: string}>()
+})
+
+test('Use `.omit` to remove a set of properties from an object', () => {
+  type Person = {name: string; age: number}
+
+  expectTypeOf<Person>().omit<'name'>().toEqualTypeOf<{age: number}>()
+})
```
mmkal, line 415: Could you move these up to come after the extract/exclude tests? And then run `pnpm lint --fix` - that will update the readme too.

mmkal (changes requested):
Thanks for this! Looks good, just need to generate the docs (eslint should do automatically based on the tests)

test/usage.test.ts
aryaemami59: Sure thing, I will do that!

README.md
aryaemami59, line 1: I believe this whole file changed due to line ending issues.

mmkal: Yeah, maybe I have something misconfigured in gitattributes. Are you on windows, out of curiosity?

aryaemami59: I'm on windows.

aryaemami59:
@mmkal  looks like snapshots needed to be updated.

mmkal (approved):

PR #56, opened 2024-03-11
Add JSDocs to everything

aryaemami59:
## This PR:

  - [X] Adds inline documentation to everything inside of the library in the form of JSDocs.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,17 +1,83 @@
+/**
+ * Negates a boolean type.
+ *
+ * @template T - The boolean type to negate.
+ */
 export type Not<T extends boolean> = T extends true ? false : true
+
+/**
+ * Returns `true` if at least one of the types in the
+ * {@linkcode Types} array is `true`, otherwise returns `false`.
+ *
+ * @template Types - An array of boolean types.
+ */
 export type Or<Types extends boolean[]> = Types[number] extends false ? false : true
+
+/**
+ * Checks if all the boolean types in the {@linkcode Types} array are `true`.
+ *
+ * @template Types - An array of boolean types.
+ */
 export type And<Types extends boolean[]> = Types[number] extends true ? true : false
+
+/**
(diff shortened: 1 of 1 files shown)


mmkal:
@aryaemami59 thanks so much, this is great! I am planning on releasing v1 very soon, so this will be great to have.

I've read through these and they look very helpful and well-written to me. I might look to tweak them/see if some of the examples can be auto-generated to make sure they stay up to date, but this is a fantastic starting point.

There's a small conflict in `src/index.ts`, could you resolve it? After that, anything else needed before marking this as ready?

aryaemami59:
@mmkal thank you, I will resolve the conflicts, not sure if there was anything else I wanted to add but I'm going to mark at as ready and if I can think of anything else to add, I'll open up a separate PR.

aryaemami59:
@mmkal Please let me know if there is anything you want me to change. I'd be happy to do it.

mmkal:
@aryaemami59 took a closer look through and some changes I think in a couple of cases these can be trimmed down a little bit. I would prefer to only have jsdoc that adds helpful, non-obvious information, rather than have 100% doc coverage. If it's repeating information that can be gleaned from the type/value itself, it risks drifting from the implementation it mirrors, as well as being noisy in the meantime. So:

1. Let's remove the `@ template T ...` tags. In I think every case, it's pretty obvious what the typearg represents
2. The docs on the symbol type helpers don't add anything IMO. `const inverted = ...`, `const expectSymbol = ...` etc. They are a quirky implementation detail for the sake of helpful error messages, there's no use case for dealing with the directly. A comment above the block that defines them along those lines would be useful though

The ones on the methods that are part of the public API are super useful. Thanks again!

aryaemami59:
@mmkal Sounds good, I will make the requested adjustments. Thanks for the feedback.

mmkal (approved):
Looks great 👍 👍 👍

PR #59, opened 2024-03-19
use eslint-plugin-mmkal

mmkal:
FYI @aryaemami59 I'm updating the lint setup to use eslint-plugin-mmkal here. It might cause minor conflicts with your PR, but any lint errors should be auto-fixable.

The new eslint-plugin-mmkal uses the flat config format so no more `mmkal/` prefix on the various eslint-disables.


aryaemami59:
@mmkal That's actually wonderful. I was going to suggest switching to flat config but you're way ahead of me. Thanks!!!

PR #61, opened 2024-03-19
Vitest

mmkal:
Another FYI @aryaemami59, I'm moving my repos gradually over from jest to vitest. Figured this would be a good one to do since vitest uses expect-type!

Adding the `vitest` dependency enables removing `jest`, `@types/jest`, `ts-jest`, `jest.config.js` and the `globals_jest` eslint config.


aryaemami59:
@mmkal That's awesome news, Vitest is the way to go.

PR #62, opened 2024-03-19
Test against different versions of TypeScript during CI

### LINKED ISSUE #60: Test against different versions of TypeScript during CI
Since we're gonna add some new features, it might be a good idea to run the tests against different versions of TypeScript during CI, it will also give us a clearer idea about which versions of TypeScript the library is compatible with. I can go ahead and put up a PR!

aryaemami59:
## This PR:

  - [X] Adds `test-types` CI job to run `tsc` with different versions of TypeScript.
  - [X] Updates `actions/checkout` usages to v4.
  - [X] Resolves #60.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,10 +16,27 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ts: ['4.7', '4.8', '4.9', '5.0', '5.1', '5.2', '5.3', '5.4']
+        node: [20.x]
+        ts: [4.7, 4.8, 4.9, 5.0, 5.1, 5.2, 5.3, 5.4]
+
     steps:
-      - uses: actions/checkout@v4
-      - run: corepack enable
-      - run: pnpm install
-      - run: pnpm install typescript@${{ matrix.ts }}
-      - run: pnpm type-check
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js ${{ matrix.node }}
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node }}
+          cache: "pnpm"
+
+      - name: Enable Corepack
+        run: corepack enable
+
+      - name: Install Dependencies
+        run: pnpm install
+
+      - name: Install TypeScript Version ${{ matrix.ts }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -1,15 +1,43 @@
...
+          cache: "pnpm"
+
+      - name: Install Dependencies
+        run: pnpm install
+
+      - name: Install TypeScript Version ${{ matrix.ts }}
+        run: pnpm install typescript@${{ matrix.ts }}
+
+      - name: Test Types
+        run: pnpm type-check
```
mmkal, line 43: I appreciate the extra docs but I am trying to slim this kind of file down as much as possible, so there's less to get my head around when I come back to it having forgotten everything a few months later. It's a bit of a personal preference thing, but I find docs that _could_ be inferred from what they're documenting a bit distracting. So how about something like this instead - I think it should work just as well:
```suggestion
      - uses: actions/checkout@v4
      - run: corepack enable
      - run: pnpm install
      - run: pnpm install typescript@${{ matrix.ts }}
      - run: pnpm type-check
```

I think each step is sufficiently self-documenting, and it makes it easier to see what the purpose of this job is at a glance.

Also, for the sake of consistency with the other job.

.github/workflows/ci.yml
```diff
@@ -1,15 +1,43 @@
...
       - run: pnpm run build
       - run: pnpm run lint
       - run: pnpm test
+
+  test-types:
+    runs-on: ubuntu-latest
+    name: Test TypeScript Types with Node.js ${{ matrix.node }} and TypeScript ${{ matrix.ts }}
+    strategy:
+      matrix:
+        node: [20.x]
```
mmkal, line 20: Let's keep the node version out of it until we want to test multiple node versions
```suggestion
    name: Test Types with TypeScript ${{ matrix.ts }}
    strategy:
      matrix:
```

mmkal (commented):
This is great!

.github/workflows/ci.yml
aryaemami59: Yes sirrr!

.github/workflows/ci.yml
aryaemami59: Your wish is my command!

Issue #55, opened 2024-03-11
`.toMatchTypeOf()` does not take into account `readonly` properties (undocumented case?)


mrazauskas:
From https://github.com/mmkal/expect-type/issues/52#issuecomment-1987805501

The documentation suggests that `.toMatchTypeOf()` can be used to "allow for extra properties". Just wanted to draw attention that it might need better description and/or examples.

Currently at least for me it sounds like `.toMatchTypeOf()` and `.toEqualTypeOf()` are identical, but `.toMatchTypeOf()` allows testing partial match. That seems to be incorrect, when `readonly` gets involved:

```ts
import { expectTypeOf } from 'expect-type';

expectTypeOf<{ readonly a: string; b: number }>().not.toEqualTypeOf<{ a: string; b: number }>();

expectTypeOf<{ readonly a: string; b: number }>().not.toMatchTypeOf<{ a: string }>(); // should also pass, but it fails
```

[playground](https://www.typescriptlang.org/play?ts=5.3.3#code/JYWwDg9gTgLgBAbzgUwB5mQYxgFQJ4YDyAZnAL5zFQQhwDkaG2AtDAcnQNwBQ3jWudiQA8SKMgCGAEwgA7ADZ44EgFxwAzjCjBZAc05wARmtkBXEIeRRyAPgAUASgB0siDCcwIAUQCOpifL4RMSiymqa2noGxnBmFla2jjx86AJByCJikjIKSqoaWjr6RibmltZk9s6u7p4AshIwmAAW6ZlhBZG6iQ4GAPR9Gs0QpvJSyvLqEHBgEurqADRGpvDA8MQSwJPcQA)

mmkal:
Discussed this elsewhere, but `expectTypeOf<A>().toMatchTypeOf<B>()` is pretty much the `A extends B` check, with a tiny amount of sanity-checking to protect against root-level `never` (and possibly `any` in future: #66).

There's an outstanding, old feature request to rename `toMatchTypeOf` to align more with typescript (i.e. something like `toExtend` maybe) and less with jest/vitest: https://github.com/mmkal/expect-type/issues/10.

But as mentioned there, since this library is already in use in a bunch of places, I want to get a v1 out before making that kind of change, or even introducing confusion with an alias.

[...]

Issue #64, opened 2024-03-19
`.toHaveProperty()` does not distinguish between `any` and `never`


mrazauskas:
[Playground link](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbwKYA8xIMYwCoE90DyAZgL5xFQQhwDkq6WAtDPkjQFDv2Y6vEA8AQwB2uAHwAKAJQA6GBAASggG5IACpXSxcEmkQgQaUgNxc0PPISL9hSVVEmz5S1RohaWu-YZOcMEYQBneAAjQSg4AF5EMkFAuFt7U3YwqBlvYyA)

`.toHaveProperty()` does not distinguish between `any` and `never`. See the link above.

It sound like `never` should not have any properties. Might be I missed something.

aryaemami59:
@mrazauskas Technically both `keyof never` and `keyof any` resolve to `string | number | symbol`. Not sure if it would be the best idea to stray away from TypeScript's intended behavior but you're right. Semantically it doesn't make sense. It seems like `keyof never` should be `never`. Not sure what the best approach here would be to be honest.

PR #65, opened 2024-03-19
Remove `test` import in `README.md`

### LINKED ISSUE #63: Add `test()` helper
[Playground link](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbwKYA8xIMYwCoE90DyAZgL5xFQQhwDkq6WAtDPkjQFCiSyJEQQAaOACMAhlDIUqtAHQyA9Hwgd2MJAGcYAChpK4LdOppCtASjgBeAHyJ2cOPPlwQogNZI46gK5QPAAyU-OAALUXV9VkRRAC44ADsvEGEkCTs4ekwcVmItJVMZGAgAWVEYDGC8QiIAHgQY+MTkiSszdjTHZzcPb184PzEoIOBw0XIvOKxgCDj9N2A4gHM4Uc0oeYXotIysSqQcgfywcVEQJDUoLQAGfMKAISQAZRg1xdb7bayqrQOZXxgfOLqGRxCAwAoQe4AQTiuFaJFMQA)

Just to draw your attention, the first example in readme is not working. I have added playground link. It is rather obvious that `test()` is not imported.

Just an idea: perhaps `expect-type` could have its own `test()` helper?

aryaemami59:
## This PR:

  - [X] Resolves #63.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,8 @@ Similar to Jest's `expect`, but with type-awareness. Gives you access to a numbe
 
 ```ts
 import {expectTypeOf} from 'expect-type'
-import {foo, bar} from '../foo'
 import {test} from 'vitest'
+import {foo, bar} from '../foo'
 
 test('foo types', () => {
   // make sure `foo` has type {a: number}
--- a/README.md
+++ b/README.md
@@ -9,17 +9,14 @@ Similar to Jest's `expect`, but with type-awareness. Gives you access to a numbe
 
 ```ts
 import {expectTypeOf} from 'expect-type'
-import {test} from 'vitest'
 import {foo, bar} from '../foo'
 
-test('foo types', () => {
-  // make sure `foo` has type {a: number}
-  expectTypeOf(foo).toMatchTypeOf<{a: number}>()
+// make sure `foo` has type {a: number}
+expectTypeOf(foo).toMatchTypeOf<{a: number}>()
 
-  // make sure `bar` is a function taking a string:
-  expectTypeOf(bar).parameter(0).toBeString()
-  expectTypeOf(bar).returns.not.toBeAny()
-})
+// make sure `bar` is a function taking a string:
+expectTypeOf(bar).parameter(0).toBeString()
+expectTypeOf(bar).returns.not.toBeAny()
 ```
 
 It can be used in your existing test files - or any other type-checked file you'd like - it's built into existing tooling with no dependencies. No extra build step, cli tool, IDE extension, or lint plugin is needed. Just import the function and start writing tests. Failures will be at compile time - they'll appear in your IDE and when you run `tsc`.
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ Similar to Jest's `expect`, but with type-awareness. Gives you access to a numbe
 ```ts
 import {expectTypeOf} from 'expect-type'
 import {foo, bar} from '../foo'
+import {test} from 'vitest'
 
 test('foo types', () => {
   // make sure `foo` has type {a: number}


mrazauskas:
Technically this is correct.

Did you notice that `foo` and `bar` are also not imported? Despite of that two assertions out of three are passing. They got type `any`. Simply imagine a situation where something `any`s because of refactoring code. `expect-type` is not able to catch such regression. (#64 is part of this problem.)

Another aspect: having Vitest import means that `tsc` will go on type checking Vitest types. Just put together a dummy repo with 2-3 test files and you will see the performance difference with import from Vitest and without. The difference is visible with naked eye. Having own `test()` helper would mean better performance for `expect-type` users.

mrazauskas:
Another limitation of `import {test} from 'vitest'` is that `test.only`, `test.skip` and similar modifiers are defined on `test()`, but do not work as expected.

Recently I saw `if()` used as dummy test helper (that library is using `tsd`):

```ts
if ('when foo and bar are imported') {
  // make sure `foo` has type {a: number}
  expectTypeOf(foo).toMatchTypeOf<{a: number}>()

  // make sure `bar` is a function taking a string:
  expectTypeOf(bar).parameter(0).toBeString()
  expectTypeOf(bar).returns.not.toBeAny()
})
```

It has no `.only` or `.skip` and does not hurt performance.

README.md
```diff
@@ -9,6 +9,7 @@ Similar to Jest's `expect`, but with type-awareness. Gives you access to a numbe
 
 ```ts
 import {expectTypeOf} from 'expect-type'
+import {test} from 'vitest'
```
mrazauskas, line 12: Hm.. If this example aims to demonstrate how `expect-type` "can be used in your existing test files", I guess following is what users would actually do:

```suggestion
import {expectTypeOf, test} from 'vitest'
```

mmkal: It's too confusing to introduce vitest this early. I'd rather get rid of the `test(...)` and just have some raw expectTypeOfs here.

mmkal:
The scope of this library is test assertions, not to be a test runner. So no need to define a `test` export.

The example in question here is just that... an example. It's to give readers an idea of what they need to do to test _their_ types. `foo` and `bar` being undefined doesn't seem important to me. Your point that two of the three assertions pass anyway because of `any` is separate, but worth considering. I've gone back and forth on whether expect-type should be strict about any even on the non-strict types like toMatchTypeOf. Currently the answer is no but maybe we should change that.

As for performance, test.skip etc.  Those are reasonable points and I think a good enough reason to remove the `test` wrapper from the first example. Nobody should install vitest _just_ for the sake of using expect-type. They can just write plain assertions. And of course installing a full-blown test runner slows down tsc (although you should try again with skipLibCheck - IMO that's always worth enabling).

Having said all that, if we remove the test wrapper from the first example, it might be nice to have a "Usage with test frameworks" section with some recommendations.

mrazauskas:
Thanks for detailed explanation.

[`Within test frameworks`](https://github.com/mmkal/expect-type?tab=readme-ov-file#within-test-frameworks) section already exists at the bottom of the readme file. Extending it indeed sounds like a good idea.

mrazauskas:
> I've gone back and forth on whether expect-type should be strict about any even on the non-strict types like toMatchTypeOf. Currently the answer is no but maybe we should change that.

Sorry about making noise here. What about documenting the above? This would explain / resolve #55 and likely #64 as well.

But documenting this behaviour will conflict with other parts of documentation, like: "first-class support for: `any` (as well as `unknown` and `never`)". Hm.. I understood that `any`s are always explicitly handled, pointers to `tsd` issues made me think so.

Also "make sure" in the following is misguiding, because `foo` can be also `any`:

```ts
// make sure `foo` has type {a: number}
expectTypeOf(foo).toMatchTypeOf<{a: number}>()
```

mmkal:
Yes, some clarification needed, how it was intended:

toEqualTypeOf: protects against any, lets you assert an exact type 
toMatchTypeOf: ensure that a value matches a type, or since typescript uses structural typing, "has" a type. That's why any passes for toMatchTypeOf.

In practice, though, you're right that very few people will _want_ an `any` except in a very small number of cases. And in those cases they could use toBeAny(). And now would be the time to make that change (pre-v1).

mrazauskas:
> As for performance, test.skip etc. Those are reasonable points and I think a good enough reason to remove the `test` wrapper from the first example. Nobody should install vitest _just_ for the sake of using expect-type. They can just write plain assertions. And of course installing a full-blown test runner slows down tsc (although you should try again with skipLibCheck - IMO that's always worth enabling).

I was making performance tests with `hyperfine`. `tsc` was running using current `tsconfig.json` where only `include` was changed to: `"include": ["test/types.test.ts", "test/usage.test.ts"]` and `"types": []` added.

With `import {test} from 'vitest'`:

- 1056 ms (average of 10 runs, with 2 warm up runs)

Replacing the import with `declare function test(name: string, fn: () => void): void;`:

- 865 ms (average of 10 runs, with 2 warm up runs)

---

Yep, `"skipLibCheck": true` is set, but it does not help.

aryaemami59:
@mmkal I removed the `test` import altogether.

PR #68, opened 2024-03-22
Document few aspects of `.toMatchTypeOf` behaviour

### LINKED ISSUE #55: `.toMatchTypeOf()` does not take into account `readonly` properties (undocumented case?)

mrazauskas:
From https://github.com/mmkal/expect-type/issues/55#issuecomment-2013750056

Closes #55
Closes #66 (perhaps?)

@mmkal Thanks for explanations. Here is an attempt to summarise of how I understood the intended behaviour of`.toMatchTypeOf`. Feel free to rework or to clarify further.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ import {expectTypeOf} from 'expect-type'
 import {foo, bar} from '../foo'
 
 // make sure `foo` has type {a: number}
-expectTypeOf(foo).toMatchTypeOf<{a: number}>()
+expectTypeOf(foo).toEqualTypeOf<{a: number}>()
 
 // make sure `bar` is a function taking a string:
 expectTypeOf(bar).parameter(0).toBeString()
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -12,7 +12,7 @@ import {expectTypeOf} from 'expect-type'
 import {foo, bar} from '../foo'
 
 // make sure `foo` has type {a: number}
-expectTypeOf(foo).toMatchTypeOf<{a: number}>()
+expectTypeOf(foo).toEqualTypeOf<{a: number}>()
```
mrazauskas, line 15: Looks like a typo.

README.md
```diff
@@ -714,4 +737,4 @@ Once you're ready to make a pull request: clone the repo, and install pnpm if yo
...
 
-After adding the tests, run `npm run lint -- --fix` to update the readme, and `npm test -- --updateSnapshot` to update the errors test. The generated documentation and tests should be pushed to the same branch as the source code, and submitted as a pull request. CI will test that the docs and tests are up to date if you forget to run these commands.
+After adding the tests, run `npm run lint -- --fix` to update the readme, and `npm test -- --update` to update the errors test. The generated documentation and tests should be pushed to the same branch as the source code, and submitted as a pull request. CI will test that the docs and tests are up to date if you forget to run these commands.
```
mrazauskas, line 740: Might be this is outdated. `--updateSnapshot` did not work for me.

README.md
```diff
@@ -145,12 +174,6 @@ expectTypeOf<never>().toBeNever()
 expectTypeOf<never>().toBeNumber()
 ```
 
-`.toEqualTypeOf` distinguishes between deeply-nested `any` and `unknown` properties:
```
mrazauskas, line 148: This is moved up.

README.md
aryaemami59: Try `--update` or just `-u`

mrazauskas: That’s right. `--update` does work as expected, hence I change this text.

aryaemami59: Yeah I think when we moved to Vitest recently we forgot to update this section of the docs.

aryaemami59 (approved):
Looks good to me, if @mmkal gives the final ok we can go ahead and merge.

PR #69, opened 2024-03-22
Fix rendering issue in readme

mrazauskas:
Currently an `eslint` comment is seeking into readme. Seems like it should not be there.


mmkal:
Thanks!

Issue #11, opened 2022-10-17
Would you mind adding a copyright / license header to index.ts?


trevorade:
Thank you for authoring this library! I'd like to start using it in my work but want to make sure that we're attributing it correctly.

Can you please add a short copyright notice to the `index.ts` file?

---

Some examples:

```ts
// Copyright 2022 [name of copyright owner]
```

```ts
/**
 * Copyright 2022 [name of copyright owner]
 * @license Apache-2.0
 */
```

mmkal:
Sorry, my reply to this didn't seem to get sent. I don't have any particular objection to this, but do you have an article/blog post explaining why this would be needed? I'd have thought the Apache license would be sufficient, and from a quick look at a random selection of other OSS repos, most don't have one baked into the actual code. Is there a tool you use that scans code for headers like this for an attribution page, or something?

trevorade:
Yeah I agree with you. Most OSS does not do this.

When pulling in open source projects into my employer's codebase, we have a policy to add a `@license` block to the JS/TS files via some automation. We also specify the copyright in these comments. I didn't see any copyright for this project though so wanted to clarify on the status there.

At this point, I don't think this matters a ton for our particular usage of `expect-type`.

mmkal:
I don't think I'll do this in that case!

PR #72, opened 2024-05-15
Update LICENSE file to properly include copyright information

trevorade:
This is related to https://github.com/mmkal/expect-type/issues/11.

I was attempting to sync this library to upstream and, again, I'm getting hung up because it doesn't include copyright information.

If you look at the end of the Apache 2 license, there are instructions on how to apply the license which I've followed here.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,18 @@
+   Copyright 2024 Misha Kaletsky
+
+   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


trevorade:
If desired, we could probably remove the rest of the license file and just include the new lines of the application of the license.

I think explicitly including the whole license after the first part is also OK.

mmkal:
Alright, I am not a lawyer but this looks fine to me!

PR #58, opened 2024-03-16
Support function overloads

aryaemami59:
## This PR:

  - [X] Adds support for function overloads through `toBeCallableWith()`
  - [x] Resolves #30 
  
  This is a rough draft, I will probably add support for overloads through `.parameter(number)` and `.parameters` as well.

Issue #30, opened 2023-05-03
support overloaded functions


y-polonsk:
Currently, the constructs like expectTypeOf.toBeCallableWith, expectTypeOf.parameters, expectTypeOf.parameter do not work with function overloads.  Basically they just assume the signature of the last overload and give error otherwise.

I believe this could be made work, doing something similar to what is described here:

https://stackoverflow.com/questions/59535995/parameters-generic-of-overloaded-function-doesnt-contain-all-options/59538756#59538756

Although, for expectTypeOf.parameter, I m not sure how to refer to a parameter at given index for specific overload.  But i believe "toBeCallableWith" and "parameters" could be made work.

Thanks.

mmkal:
Oh that's an interesting workaround. Yes, I'd be open to adding that. Let me know if you would like to create a PR. There are some big-ish perf changes in flux right now, but this would probably be separate enough that it doesn't affect them.

aryaemami59:
@mmkal I'm gonna give this a shot!

PR #83, opened 2024-07-26
improve overloads support, attempt 2

mmkal:
Related to #58
Related to #30

Improve support for overloaded functions (up to 10 overloads). So for an example function type:

```ts
type Factorize = {
  (input: number): number[]
  (input: bigint): bigint[]
}
```

---

`.parameters` gives you an ExpectTypeOf instance with a union of the parameter-tuple types, so:
 
```ts
expectTypeOf<Factorize>().parameters.toEqualTypeOf<[number] | [bigint]>()
```

---

`.returns` gives you an ExpectTypeOf instance with a union of the return types, so

```ts
expectTypeOf<Factorize>().returns.toEqualTypeOf<number[] | bigint[]>()
```

---

`.toBeCallableWith(...)` now accepts any overload input, not just the "last" like before. And you can now chain it via `.returns` which gives you the matching return type:

```ts
expectTypeOf<Factorize>().toBeCallableWith(5).returns.toEqualTypeOf<number[]>()
```

---

## Implementation

- overload utilities were added
   - initially, I thought all that was needed was a utility that matches `F` typeargs against a single 10-overload type
   - but the `test-types` job caught that this _doesn't_ work for typescript <5.3
   - for lower typescript versions, it seems we need an approach more like the one in #58
   - there are some edge cases the tests found for generic functions and parameterless functions, so there is _sometimes_ some "useless" overload info that has to be explicitly excluded
   - there are a couple of intermediate utilities to check if the 10-overload type can be used, and to exclude "useless" overloads
- equivalents added for constructor parameters too
- used the overloaded versions in `DeepBrand` too

<details>
<summary>update: moved these changes to #93 to reduce diff</summary>

- I felt the megafile `index.ts` was finally getting too big with the added overload utilties:
  - so those were put in a new `overloads.ts`
  - since overload utils they rely on some other existing utils, I created `utils.ts` to avoid circular references
  - `index.ts` remains as the home for the main `ExpectTypeOf` exports
  - added `export * from './utils'` since we have been exporting all the internal utils and I don't want to break people
  - update: moved these changes to #93 to reduce diff
</details>


aryaemami59:
Sorry I took too long with #58. Does this supersede that one?

mmkal:
Yes, I wrote this using that as a reference, I'd like it to be a replacement yeah. Need to go through carefully and make sure nothing is dropped between the two though.

aryaemami59:
Gotcha, I will look through it soon.

README.md
```diff
@@ -597,6 +599,19 @@ class B {
 // Instead of the above, try something like this:
 expectTypeOf(B).instance.toEqualTypeOf<{b: string; foo: () => void}>()
 ```
+
+Overloads limitation: overloaded functions which include an overload looking like `(...args: unknown[]) => unknown` will excluded `unknown[]` from `.parameters` and exclude `unknown` from `.returns`. This overload just allows any input and returns unknown output anyway though.:
```
aryaemami59:
```suggestion
Overloads limitation: Overloaded functions that include an overload resembling `(...args: unknown[]) => unknown` will exclude `unknown[]` from `.parameters` and `unknown` from `.returns`. This overload, however, allows any input and returns an unknown output anyway:
```

README.md
```diff
@@ -742,6 +757,10 @@ The key differences in this project are:
 - built into existing tooling. No extra build step, cli tool, IDE extension, or lint plugin is needed. Just import the function and start writing tests. Failures will be at compile time - they'll appear in your IDE and when you run `tsc`.
 - small implementation with no dependencies. [Take a look!](./src/index.ts) (tsd, for comparison, is [2.6MB](https://bundlephobia.com/result?p=tsd@0.13.1) because it ships a patched version of typescript).
 
+## TypeScript backwards-compatibility
+
+There is a CI job `test-types` which checks that the tests still pass with certain older typescript versions. To check supported typescript versions, [look at the job definition](./.github/workflows/ci.yml).
```
aryaemami59, line 768:
```suggestion
There is a CI job called `test-types` that checks whether the tests still pass with certain older TypeScript versions. To check the supported TypeScript versions, [refer to the job definition](./.github/workflows/ci.yml).
```

README.md
```diff
@@ -751,3 +770,7 @@ Once you're ready to make a pull request: clone the repo, and install pnpm if yo
...
 
 After adding the tests, run `npm run lint -- --fix` to update the readme, and `npm test -- --updateSnapshot` to update the errors test. The generated documentation and tests should be pushed to the same branch as the source code, and submitted as a pull request. CI will test that the docs and tests are up to date if you forget to run these commands.
+
+### Documentation of limitations through tests
+
+Limitations of the library are documented via tests in usage.test.ts. This means that if a future TypeScript version (or library version) fixes the limitation, the test will start failing, so it will be automatically removed from the docs once it no longer applies.
```
aryaemami59, line 782:
```suggestion
Limitations of the library are documented through tests in `usage.test.ts`. This means that if a future TypeScript version (or library version) fixes the limitation, the test will start failing, and it will be automatically removed from the documentation once it no longer applies.
```

test/usage.test.ts
```diff
@@ -446,3 +446,14 @@ test('Another limitation: passing `this` references to `expectTypeOf` results in
   // Instead of the above, try something like this:
   expectTypeOf(B).instance.toEqualTypeOf<{b: string; foo: () => void}>()
 })
+
+test('Overloads limitation: overloaded functions which include an overload looking like `(...args: unknown[]) => unknown` will excluded `unknown[]` from `.parameters` and exclude `unknown` from `.returns`. This overload just allows any input and returns unknown output anyway though.', () => {
```
aryaemami59:
```suggestion
test('Overloads limitation: Overloaded functions that include an overload resembling `(...args: unknown[]) => unknown` will exclude `unknown[]` from `.parameters` and `unknown` from `.returns`. This overload, however, allows any input and returns an unknown output anyway.', () => {
```

mmkal: I removed this in the end/ replaced it with a similar hand-written part of the readme. It's only a limitation that applies on old typescript versions.

README.md
mmkal: re-jigged this a little since you commented but done.

mmkal:
@aryaemami59 I've updated the description of this pull request to explain what's going on in it better, both in terms of the API surface change, and the implementation. Worth having a look at if you're planning on diving into the code.

Would be great to get your thoughts on whether you think this is a good idea, or whether it's too much complexity.

I think it's quite possible there are more edge cases around generics etc., and it could get out of hands if there's more edge case handling needed. On the other hand, kind of the point of this library is to abstract that kind of complexity away from regular typescript users who just want to test their types in a sensible way, so maybe the more complex it is, the more value this library is bringing - even if it does become harder to maintain.

CC @trevorade and @y-polonsk in case you want to weigh in too.

aryaemami59:
@mmkal, I'm going to take a deep dive today and will let you know my full thoughts on it.

Here are my initial thoughts:

- I really like that you can now chain `toBeCallableWith()` with `.returns`. I think this is something we should explore and possibly use more often.
- If `.parameters` now requires a union of parameter-tuple types and `.returns` requires a union of return types, does this mean it will cause a breaking change? If so, do you think this is something we should add as an opt-in feature?

package.json
```diff
@@ -40,6 +41,7 @@
     "np": "^10.0.0",
     "strip-ansi": "6.0.1",
     "ts-morph": "16.0.0",
+    "ts52": "npm:typescript@5.2.2",
```
aryaemami59, line 44: Since we have the `test-types` CI job, do we need this?

aryaemami59: Don't worry about it I can remove it, hope you don't mind :)

sheremet-va:
Not sure what contribution I can make here, but it looks good to me 👍🏻

aryaemami59:
@mmkal

First, let me express my sincere gratitude for the detailed explanations in the JSDocs. Explicit inline documentation is something I deeply appreciate (though I understand it's not everyone's preference). That said, I spent the entire morning running this PR against a variety of type tests, some of which I was very familiar with (OSS projects I've contributed to before), while others were entirely new to me. Overall, the results were pretty good, with far less breakage than I initially anticipated. I do, however, have a few questions and observations:

- Previously, because `toBeCallableWith()` didn't support overloads, some users resorted to workarounds like using `.parameter(0).toEqualTypeOf()` or `.parameters.toEqualTypeOf()`. This is where most of the breakage I mentioned earlier seems to originate. Would it be possible to support overloads without breaking these tests? seems like the changes to `.parameters` and `.returns` could be it's own feature. Another thing that concerns me is that the `.parameters`/`.returns` changes could lead to a mismatch in semantics. Previously, many methods mirrored their TypeScript counterparts closely (e.g., `.parameters` corresponding to `Parameters<>` and `.parameter(0)` corresponding to `Parameters<>[0]`). Now, each usage may need to be treated differently depending on whether the function has multiple signatures. It could make things more inconsistent. I'm not entirely convinced that the benefits outweigh the drawbacks in this case.

- Is there a way to determine the number of overloads a function has? From my understanding, this isn't really feasible before TypeScript v5.3, but I'd appreciate your insights.

- Now that `.returns` can be chained after `.toBeCallableWith()`, do you think it would be beneficial or even possible to allow chaining of `.parameter()` and `.parameters` as well?

- How are negated assertions impacted by these changes?

I'm definitely going to have more questions later once I take another look and most likely write some unit tests. In the meantime you can also let me know if there is anything else I can do, and thank you for your hard work on this.

mmkal:
>do you think it would be beneficial or even possible to allow chaining of .parameter() and .parameters as well?

Yes, good idea, .parameter and .parameters would be useful too.

>Previously, many methods mirrored their TypeScript counterparts closely

It is true that things like pick and omit correspond pretty much exactly to typescript builtins, but I think it's ok that in certain cases expect-type goes beyond builtin behaviour, where it can take advantage of its scope being more narrow than typescript's. I'm still struggling to think of a case where it'd be preferable to ignore all but one overload. Examples would be helpful if you have some (maybe the breakages you referred to), and for my part I'll look for the typescript team's reasoning for why the builtins work the way they do, and see whether it applies here too.

Having said that, this should be called out more explicitly, and we should link to the various typescript issues where they explain why Parameters<...> et al. just pick the last overload.

>Is there a way to determine the number of overloads a function has?

I think this was actually more possible pre 5.3. Now, any function matches a 10-overload type, so the only way to count would be to equality check to eliminate duplicates. So I don't know if it's a good idea to add that, but if you find something different I'm open to it.

>How are negated assertions impacted by these changes?

Hopefully they're not really. .not.toBeCallableWith(...) is already disallowed and .parameters.not.toEqualTypeOf<[number]>() should still work as expected. Can add a test for this though.

aryaemami59:
@mmkal Mind if I take a crack at chaining of `.parameters` and `.parameter()`?

mmkal:
I have a local branch with it! One last thing I need to improve with it, but if I have trouble would be great to get your help. Will let you know.

aryaemami59:
Awesome thanks!

mmkal:
Updates since last time:

- convert overloads to a union of functions rather than a tuple
- various simplifications that fall out of that
- constructor parameters support in a similar way
- some non-readme tests in types.test.ts

Next:

- use the overload helpers in `DeepBrand` - currently it won't distinguish between overloaded functions and we should be consistent

src/overloads.ts
```diff
@@ -0,0 +1,182 @@
...
+ */
+export type TSPost53OverloadsInfoUnion<F> =
+  F extends {(...args: infer A1): infer R1; (...args: infer A2): infer R2; (...args: infer A3): infer R3; (...args: infer A4): infer R4; (...args: infer A5): infer R5; (...args: infer A6): infer R6; (...args: infer A7): infer R7; (...args: infer A8): infer R8; (...args: infer A9): infer R9; (...args: infer A10): infer R10}
+  ? ((...p: A1) => R1) | ((...p: A2) => R2) | ((...p: A3) => R3) | ((...p: A4) => R4) | ((...p: A5) => R5) | ((...p: A6) => R6) | ((...p: A7) => R7) | ((...p: A8) => R8) | ((...p: A9) => R9) | ((...p: A10) => R10)
+  : never
+
+export type UnknownFn = (...p: unknown[]) => unknown
+
+/**
+ * `true` iff `T` is equivalent to `(...p: unknown[]) => unknown`, which is what an overload variant looks like for a non-existent overload
```
aryaemami59, line 18:
```suggestion
 * `true` if `T` is equivalent to `(...p: unknown[]) => unknown`, which is what an overload variant looks like for a non-existent overload.
```

src/overloads.ts
```diff
@@ -0,0 +1,182 @@
...
+ * This is useful because older versions of TypeScript end up with 9 "useless" overloads and one real one for parameterless/generic functions
+ */
+export type IsUselessOverloadInfo<T> = StrictEqualUsingTSInternalIdenticalToOperator<T, UnknownFn>
+
+/**
+ * Old versions of typescript can sometimes seem to refuse to separate out union members unless you put them each in a pointless tuple and add an extra `infer X` expression.
```
aryaemami59, line 24:
```suggestion
 * Old versions of TypeScript can sometimes seem to refuse to separate out union members unless you put them each in a pointless tuple and add an extra `infer X` expression.
```

aryaemami59 (changes requested):

src/overloads.ts
mmkal: `iff` wasn't a typo - it means "if and only if". With just if it'd allow for it to be true for other types too

src/overloads.ts
```diff
@@ -0,0 +1,93 @@
...
+export type TSPost53OverloadsInfoTuple<F> =
+  F extends {(...args: infer A1): infer R1; (...args: infer A2): infer R2; (...args: infer A3): infer R3; (...args: infer A4): infer R4; (...args: infer A5): infer R5; (...args: infer A6): infer R6; (...args: infer A7): infer R7; (...args: infer A8): infer R8; (...args: infer A9): infer R9; (...args: infer A10): infer R10}
+  ? [{parameters: A1; return: R1}, {parameters: A2; return: R2}, {parameters: A3; return: R3}, {parameters: A4; return: R4}, {parameters: A5; return: R5}, {parameters: A6; return: R6}, {parameters: A7; return: R7}, {parameters: A8; return: R8}, {parameters: A9; return: R9}, {parameters: A10; return: R10}, ]
+  : never
+
+export type BaseOverloadInfo = {parameters: unknown[]; return: unknown}
+
+/**
+ * `true` iff `T` is equivalent to `{parameters: unknown[]; return: unknown}`, which is what an overload info object looks like for a non-existent overload
+ * This is useful because older versions of TypeScript end up with 9 "useless" overloads and one real one for parameterless/generic functions
```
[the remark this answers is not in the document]
mmkal: More that old versions of typescript seem "fill up" overload slots with `(..._: unknown[]) => unknown` functions. I'll try to make a playground for ts5.2 and link it here so it's clearer.

src/overloads.ts
```diff
@@ -0,0 +1,93 @@
...
+ */
+export type IsUselessOverloadInfo<T> = StrictEqualUsingTSInternalIdenticalToOperator<T, BaseOverloadInfo>
+
+// prettier-ignore
```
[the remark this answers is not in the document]
mmkal: I'd rather keep them because otherwise prettier makes these types extremely long - it always puts function overloads on individual lines. Better IMO to have one type per line since there's very little interesting information in the explicit counting from 1 to 10

src/index.ts
```diff
@@ -515,7 +516,16 @@ export interface BaseExpectTypeOf<Actual, Options extends {positive: boolean}> {
    * @param args - The arguments to check for callability.
    * @returns `true`.
    */
-  toBeCallableWith: Options['positive'] extends true ? (...args: Params<Actual>) => true : never
+  toBeCallableWith: Options['positive'] extends true
+    ? <A extends OverloadParameters<Actual>>(
+        ...args: A
+      ) => {
+        parameters: ExpectTypeOf<FilterExtendedBy<OverloadParameters<Actual>, A>, Options>
+        parameter: <N extends number>(n: N) => ExpectTypeOf<FilterExtendedBy<OverloadParameters<Actual>, A>[N], Options>
```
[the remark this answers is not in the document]
mmkal: Sadly not, that was a what I tried first. That still allows any number input, even for tuples, so not sure there's a point. There might be a ts issue for it somewhere.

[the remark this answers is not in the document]
mmkal: Ok yeah it's that any number key is allowed, but string keys need are properly type-checked. We could consider making this a string type instead, and use template literals to only allow number-like strings, but I don't know if it's worth it...

image

src/utils.ts
```diff
@@ -364,3 +364,14 @@ export type Scolder<
...
+
+/**
+ * Keeps the types in a union which are extended by `T` - sort of similar to `Extract` except the other way around.
+ *
+ * @example
+ * ```ts
+ * type MyOverloadParameters = [number] | [string] | [boolean]
+ *
+ * type FilteredOverloadParameters = FilterExtendedBy<MyOverloadParameters, [6]> // [number]
+ */
```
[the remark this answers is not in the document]
mmkal: removed

src/index.ts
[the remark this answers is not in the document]
mmkal: I'm going to leave this as is because it's not bad anyway - if you do `.parameter(1.5)` the follow-on assertion would fail (like `.parameter(0.5).toBeString()` - the only one that would pass would be `.parameter(0.5).toBeUndefined()` which matches runtime behaviour anyway.

If you can find a way to make it more strongly type-checked I'd be open to a follow-on PR but this one already is getting too big.

src/overloads.ts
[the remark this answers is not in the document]
mmkal: it's a bug fixed in ts 5.3. [Simplified demo](https://www.typescriptlang.org/play/?ts=5.2.2#code/C4TwDgpgBAIhDGAnCBDAzgSwHYHMDyAbhIgDYD2KAJmgJJYBmZAqlhmVgDwBiAfFALxQuUCAA9gELNSgBvABQA6JSkQ40ALijZ6xKAEEAjAEpN23QCUDAbiiLlqjVoa69AJhNOdiKOdc27Cipqps7eegDMHmbe5uH+SoEOIV76ACxRoT6p8fbBni4ArBkp5gU5iXnR+gBsxRbV5UGOVXoA7HUxrY1J+WEAHB0+fd2VmXoAnIPm4yPNYwYADFOLAL4AUFBQAPy2AWCahkYCfJZHAD67Cfv67sc+txdyeweRd7Hnl0rXeulvv4-PfRFN7AgFXA61N6QsFfA7tN7wmEKb4DN6opHfSZvLEYg6LI78E74jZQTTCMQSKRoWQBJrJFzGekxay2BJ03o3KZ+Vm5OYpCJTOI8ip8lzpJlZWYSvRFCWlKUcvS1OUNYXslrtOVdNU9FoDOXDHWjfmTOUzKDrTabHZPcH6YxvB24zlvB6fZEvAknV7On5erIfW2woH+0qBwFK0PQ93feGEnyImMHVHx8zopP6LGpyYkzZkkTiSTSeRs3XzZYKlruOXc2ll-mROVCuvGsVTbJG0VhWUc+Wd6XK3uqltd-Sa3vakfS-W9w2Wq3WjOHR3hu1uUNuoMe-SvVM+pe-VP-JfA1OgpeQ1PRrex0OJm-J0MDXNW-MUovUku86WM3ssqeKtWva1qWrZhI2vbNqBo5+nKHYAS0PZVH2CFjIOyHDtB0rjsh2rzgui4PvaoZOkutyppuEa7t6q7Bn6fy0duMqhueRGRlCjG3giRgvgub6FlSNJYYqv7If+wlVlylZjBByFQd+irir28ESWMSGZChqn8uhGmqvhBGERGDqpqRbHkSclFrtRPj7mxh4nMebGnicrERpeJy1LxBH8ZSxaofyokaeJCmSTW0kNoK4VtnBUXdlMZQWl5Bk2kZJGcQc5n3OlO6hrZEb2QGUC+s5PhFEl3lCAWvmfv5DIVv2gFSQ1LSyRp8kitKSnIR2+kGQuKVrsZRLZeurojdZ7xFQeobpOVBk+R+QkheWcrBR1jVhc1MmRYlfV7dsS5DT4pkRplvjjblPH7X1C2CV+60tIFJRreqYxAch3K9ddVoDXRR2nFNZkbld33zZV76CbVYRHWJB1sf9xhzd9mhYBARCIGsaygJAUDAAYAiwAgyDoNg+Do+QVC0AwzCsOwHDyCgmi-v+jNQEBrgrHwAD0XO4+A0B4wTTys0dJ2s5l7iY9jAuuATcBIKgmC4IQxAU9QdCMCwbCcHI-oGNzvPS7jsuCERACuWAANZYGQADuWAANoALr+hb1t21g2Vuzb9vO67Vs+57gOAt7Ht+3cof217Adhy7Ecx1Hwd2pHjtx-GKfR+7vtp3wGdJ8GKfh+nCdB86hc51AedlyXRe5yX4Z6zxQA)

This is the bug, I added it to the jsdoc for `TSPre53OverloadsInfoUnion` but will add it to here too: https://github.com/microsoft/TypeScript/issues/28867

aryaemami59 (approved):

aryaemami59:
Alright let's get this in.

PR #87, opened 2024-08-01
setup `deps` branch + renovate config

mmkal:
CC @aryaemami59 ~going to~ hoping to try to something I did in https://github.com/mmkal/trpc-cli to reduce renovate noise.

Some explanaition of what/why: https://x.com/mmkalmmkal/status/1806690658456862998

Here's an example PR on trpc-cli that's got two weeks worth of dependency updates, available as a single squashable, green, pull request: https://github.com/mmkal/trpc-cli/pull/30

I'll hold off til 1) it's green and 2) to give you a chance to raise any objections, this is not exactly a super tried and tested strategy, and there might be some fatal flaw in it that I've missed so far in trpc-cli

### DIFF
--- a/.github/workflows/deps.yml
+++ b/.github/workflows/deps.yml
@@ -1,0 +1,116 @@
+name: deps-sync
+on:
+  push:
+    branches:
+      - main
+      - deps
+      - deps-sync
+      - 'deps-sync/**'
+
+jobs:
+  run:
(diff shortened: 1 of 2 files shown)


aryaemami59:
As of now, I don't have any objections to this. I'd say let's try it, and if it somehow becomes problematic, we can revert back.

PR #93, opened 2024-08-07
create utils file

mmkal:
this is mostly to reduce the unrelated diff on https://github.com/mmkal/expect-type/pull/83 - it's essentially the same refactor of moving the type-utils like `Not<...>`, `Extends<...>`, `Eq<...>` et al. to a utils.ts file. I would like for that to be strictly overloads-related as much as possible esp. since it's a potentially controversial change.

FYI @aryaemami59 not sure if this will make it harder for you to merge main into your other PR, or if you plan on continuing to do that.


aryaemami59:
This is wonderful. Thanks for the heads up.

PR #95, opened 2024-08-09
branding.ts and messages.ts

mmkal:
Another intermediate PR in support of #83

Moves `DeepBrand` to its own file, because in #83 I want to start using imports from `overloads.ts` in it. This would be a circular reference, which TypeScript can actually handle, but I would still like to avoid to prevent confusion. 

Similarly I had to create _another_ file `messages.ts` since we have a few type-error aiding utilities that are currently in utils.ts, but that rely on `DeepBrand`.

This tends to be the kind of file-proliferation that I worry about when breaking up a mega-file! But in this case I think it's warranted.

@aryaemami59 what do you think re naming/ the new structure?


src/utils.ts
```diff
@@ -238,22 +129,10 @@ export type StrictEqualUsingTSInternalIdenticalToOperator<L, R> =
...
- * Represents a type that checks if two types are equal, using
- * a hopefully performant approach.
- * It first checks if the types are strictly equal using
- * {@linkcode StrictEqualUsingTSInternalIdenticalToOperator}.
- * If they are not strictly equal, it falls back to using the
- * {@linkcode StrictEqualUsingBranding} type.
+ * Checks that `L` and `R` extend each other. Not quite the same as an equality check since `any` can make it resolve to true.
+ * So should only be used when `L` and `R` are known to avoid `any`.
  */
-export type HopefullyPerformantEqual<Left, Right> =
```
mmkal, line 255: Removed because it's unused. This was a type that I think I originally intended to avoid the need to add `.branded` manually for types that were tripped up by intersections, but I think I remember it not working reliably so am removing. Highly unlikely it's used by anyone.

aryaemami59:
Only minor issue I might have is instead of `brand.ts` maybe it can be called `branding.ts` or maybe `branded-types.ts` other than that LGTM.

PR #97, opened 2024-08-10
Add `property` method to extract object properties

mikoloism:
Hello @mmkal! 👋

Thank you for your hard work on the `expect-type` library! I really appreciate the effort you put into maintaining such a valuable tool for the community.

As I’ve been using the library, I noticed that it could be even more powerful with the addition of a new method. This pull request proposes the addition of a `property` method that allows users to extract specific properties from an object and perform type assertions on them. 

The `property` method will enable users to easily check the types of individual properties within an object, enhancing the library's functionality and improving its usability.

For example:

```typescript
const user = {
  name: 'Alice',
  age: 30,
};

expectTypeOf(user).property('name').toBeString();
expectTypeOf(user).property('age').toBeNumber();
```

Let me know if you have any questions or need further changes!

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,6 +271,43 @@ type Person = {name: string; age: number}
 expectTypeOf<Person>().pick<'name'>().toEqualTypeOf<{name: string}>()
 ```
 
+Use `.property` to pick a property from an object and narrow down complex objects:
+
+```typescript
+type CSSDisplay = 'flex' | 'grid' | 'block';
+interface CSSProperties {
+  display: CSSDisplay | `inline-${CSSDisplay}`;
+  font?: {
+    size?: 'lg' | 'sm' | 'xl' | 'base';
+    style?: 'normal' | 'italic' | 'bold';
+  };
+}
+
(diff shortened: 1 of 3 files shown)


aryaemami59:
Have you tried the `toHaveProperty` method that we have? It seems like they might be very similar if not identical.

mikoloism:
> Have you tried the `toHaveProperty` method that we have? It seems like they might be very similar if not identical.

I’ve had a moment of clarity and realized I might have been a bit too eager with my pull request. 😅 I apologize for the oversight regarding the `toHaveProperty` method!

I’ll go ahead and close this pull request. Thank you for your understanding, and keep up the fantastic work on the library!

mmkal:
Closing this for now - thank you for taking the time to create a PR! And if there's anything you think could be added to the docs to make it clearer that you can use `.toHaveProperty(...)` in this way, please feel free to submit another one (this might be in the form of a usage.test.ts change which updates as you've seen in this PR).

PR #98, opened 2024-08-12
.map fn

mmkal:
@aryaemami59 another potential helper that occurred to me might be useful. I'm not totally convinced it's worth it, but thought I'd open a draft PR in case you had thoughts. Given `.toBeCallableWith` doesn't work well for generic functions, it's a catch-all way to say "I am allowed to do this with my object" and make assertions on the result, without actually calling the function or performing side-effects, etc.:

```ts
const capitalize = <S extends string>(input: S) =>
  (input.slice(0, 1).toUpperCase() + input.slice(1)) as Capitalize<S>

expectTypeOf(capitalize)
  .map(fn => fn('hello world'))
  .toEqualTypeOf<'Hello world'>()
```

But the assertion could just as easily (maybe _more_ easily) be written as:

```ts
expectTypeOf(() => capitalize('hello world')).returns.toEqualTypeOf<'Hello world'>()
```

Which also doesn't actually call `capitalize` at runtime.

Advantages of `.map`:

- it makes it clearer it's an asertion about `capitalize` rather than an inline callback.
- it works fine with `expectTypeOf<X>()` as well as `expectTypeOf(x)`
- implementation is v simple so the cost is low

It could also sub in for how we do `.toBeCallableWith` in #83 too:

```ts
type Factorize = {
  (input: number): number[]
  (input: bigint): bigint[]
}

expectTypeOf<Factorize>().map(f => f(12)).toEqualTypeOf<number[]>()
expectTypeOf<Factorize>().map(f => f(12n)).toEqualTypeOf<bigint[]>()
```


aryaemami59:
Not sure about this one yet, I'd say let's get #83 in first then we can use this one to potentially fill in any left over gaps.

mmkal:
@aryaemami59 I keep forgetting this is still a draft. Not a particularly exciting feature but v small implementation cost and somewhat useful so I'd like to get it in. Let me know if you have any thoughts on the code.

src/index.ts
```diff
@@ -499,6 +499,14 @@ export interface BaseExpectTypeOf<Actual, Options extends {positive: boolean}> {
    */
   toBeNullable: Scolder<ExpectNullable<Actual>, Options>
 
+  /**
+   * Transform that type of the value via a callback.
+   *
+   * @param fn A callback that transforms the input value. Note that this function is not actually called - it's only used for type inference.
```
aryaemami59, line 505:
```suggestion
   * @param fn - A callback that transforms the input value. Note that this function is not actually called - it's only used for type inference.
```

aryaemami59 (changes requested):

aryaemami59 (approved):
Still not too sure about the name tbh, and we'll add more docs and examples later to cover some of the niche use cases. Other than all looks good to me.

PR #108, opened 2024-08-15
Add `are-the-types-wrong` CI job

### LINKED ISSUE #105: Add `are-the-types-wrong` CI job
It might be worth adding a CI job that runs [Are-the-types-wrong](https://github.com/arethetypeswrong/arethetypeswrong.github.io) to check the final type definitions.

aryaemami59:
## **This PR**:

- [X] Add `are-the-types-wrong` CI job.
- [X] Resolves #105.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,13 +8,9 @@ jobs:
       - uses: actions/checkout@v4
       - run: corepack enable
       - run: pnpm install
+      - run: pnpm run build
       - run: pnpm run lint
       - run: pnpm test
-      - run: mv $(pnpm pack | tail -n1) pkg.tgz
-      - uses: actions/upload-artifact@v4
-        with:
-          name: artifact
-          path: pkg.tgz
 
   test-types:
     runs-on: ubuntu-latest
@@ -31,11 +27,9 @@ jobs:
 
   are-the-types-wrong:
     runs-on: ubuntu-latest
-    needs: run
     steps:
+      - uses: actions/checkout@v4
       - run: corepack enable
-      - uses: actions/download-artifact@v4
-        with:
-          name: artifact
-          path: ./
-      - run: pnpx @arethetypeswrong/cli@latest ./pkg.tgz
+      - run: pnpm install
+      - run: pnpm pack
+      - run: pnpx @arethetypeswrong/cli@latest $(ls | grep tgz)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,9 +33,6 @@ jobs:
     runs-on: ubuntu-latest
     needs: run
     steps:
-      - uses: actions/checkout@v4
-      - run: corepack enable
-      - run: pnpm install
       - uses: actions/download-artifact@v4
         with:
           name: artifact
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,8 +4,6 @@ on: [push, pull_request]
 jobs:
   run:
     runs-on: ubuntu-latest
-    outputs:
-      artifact: ${{ steps.pack.outputs.tarball }}
     steps:
       - uses: actions/checkout@v4
       - run: corepack enable
@@ -16,7 +14,7 @@ jobs:
       - uses: actions/upload-artifact@v4
         with:
           name: artifact
-          path: ./*.tgz
+          path: pkg.tgz
 
   test-types:
     runs-on: ubuntu-latest
@@ -41,5 +39,5 @@ jobs:
       - uses: actions/download-artifact@v4
         with:
           name: artifact
-          path: ./
-      - run: pnpx @arethetypeswrong/cli@latest ${{ needs.run.outputs.artifact }}
+          path: pkg.tgz
+      - run: pnpx @arethetypeswrong/cli@latest pkg.tgz
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: run
     steps:
+      - run: corepack enable
       - uses: actions/download-artifact@v4
         with:
           name: artifact
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,8 +12,7 @@ jobs:
       - run: pnpm install
       - run: pnpm run lint
       - run: pnpm test
-      - id: pack
-        run: echo "tarball=$(pnpm pack | tail -n1)" >> $GITHUB_OUTPUT
+      - run: mv $(pnpm pack) pkg.tgz
       - uses: actions/upload-artifact@v4
         with:
           name: artifact
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
       - run: pnpm install
       - run: pnpm run lint
       - run: pnpm test
-      - run: mv $(pnpm pack) pkg.tgz
+      - run: mv $(pnpm pack | tail -n1) pkg.tgz
       - uses: actions/upload-artifact@v4
         with:
           name: artifact
@@ -40,4 +40,4 @@ jobs:
         with:
           name: artifact
           path: pkg.tgz
-      - run: pnpx @arethetypeswrong/cli@latest pkg.tgz
+      - run: pnpx @arethetypeswrong/cli@latest ./pkg.tgz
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -24,3 +31,16 @@ jobs:
...
       - run: pnpm install typescript@${{ matrix.ts }}
       - run: pnpm type-check
+
+  are-the-types-wrong:
+    runs-on: ubuntu-latest
+    needs: run
+    steps:
+      - uses: actions/checkout@v4
+      - run: corepack enable
+      - run: pnpm install
```
mmkal, line 41: Are these necessary? They might actually hide errors eg if we accidentally use types from node_modules

.github/workflows/ci.yml
```diff
@@ -4,13 +4,20 @@ on: [push, pull_request]
...
+      artifact: ${{ steps.pack.outputs.tarball }}
     steps:
       - uses: actions/checkout@v4
       - run: corepack enable
       - run: pnpm install
-      - run: pnpm run build
       - run: pnpm run lint
       - run: pnpm test
+      - id: pack
+        run: echo "tarball=$(pnpm pack | tail -n1)" >> $GITHUB_OUTPUT
```
mmkal, line 16: I think we can simplify
```suggestion
      - run: mv $(pnpm pack) pkg.tgz
```

Then we don't need to export it as an output of this step or job, or use *.tgz below, we just know it's called pkg.tgz

.github/workflows/ci.yml
```diff
@@ -24,3 +31,16 @@ jobs:
...
+    needs: run
+    steps:
+      - uses: actions/checkout@v4
+      - run: corepack enable
+      - run: pnpm install
+      - uses: actions/download-artifact@v4
+        with:
+          name: artifact
+          path: ./
+      - run: pnpx @arethetypeswrong/cli@latest ${{ needs.run.outputs.artifact }}
```
mmkal, line 46: A thought: I wonder if attw could work using a custom registry? Arguably this fits better in the pkg.pr.new workflow. 

I know you can do `attw --from-npm expect-type` but maybe it'd work with the pkg.pr.new url too. What do you think?

package.json
```diff
@@ -31,6 +31,7 @@
     "lint": "tsc && pnpm eslint .",
     "type-check": "tsc",
     "build": "tsc -p tsconfig.lib.json",
+    "prepack": "pnpm run build",
```
mmkal, line 34: Nit:
```suggestion
    "prepack": "pnpm build",
```

.github/workflows/ci.yml
aryaemami59: Worth a try, though `pkg.pr.new` has not been kind to me.

.github/workflows/ci.yml
aryaemami59: How would we accidentally use types from `node_modules`?

.github/workflows/ci.yml
```diff
@@ -4,13 +4,19 @@ on: [push, pull_request]
 jobs:
   run:
     runs-on: ubuntu-latest
+    outputs:
+      artifact: ${{ steps.pack.outputs.tarball }}
     steps:
       - uses: actions/checkout@v4
       - run: corepack enable
       - run: pnpm install
-      - run: pnpm run build
```
mmkal, line 11: I don't love that this is being removed - the purpose of this job is validate the project is set up right, and the build step is part of that. Sure, it's now part of pnpm pack, which is a good new thing, but it's not very visible.

aryaemami59: I can separate them

.github/workflows/ci.yml
mmkal: I'm actually not sure how attw checks so maybe it's not possible. 

But since installing here would add a node_modules folder, an import in the package would be able to resolve stuff from node_modules, which we wouldn't want except for prod dependencies - and we don't have any.

.github/workflows/ci.yml
mmkal: Oh interesting what have you seen?

.github/workflows/ci.yml
aryaemami59: Just realized by re-adding `npm run build` technically we run the `build` script twice.

.github/workflows/ci.yml
aryaemami59: Screenshot (80)

.github/workflows/ci.yml
aryaemami59: `attw` inspects the tarball and checks to see if the type definitions are valid based on `package.json`'s `types` and `exports` fields (oversimplifying).

mmkal: Yeah, I think my comment got shifted somehow but what I'm objecting to is checkout/ corepack enable/ pnpm install - even if they doesn't add the risk I suggested they are still unnecessary if we're using pnpx, as far as I can tell.

aryaemami59: Oh I see, I misunderstood. Okay let me try fixing it.

aryaemami59: `corepack enable` turned out to be necessary but I got rid of the other stuff.

.github/workflows/ci.yml
mmkal: I suspect I set it up in a way that makes it not-work from forked repos. (Side-note, you should have permissions to push non-protected branches to mmkal/expect-type, but probably good that you're not doing that yet to help debugging this!)

If we can get it working, I think it might be a better option. I just tried it out with the pkg.pr.new link + curl and it works v easily:

image

So this whole thing could reduce down to a two-liner if it's possible to fix up that pkg-pr-new workflow.

What do you think?

mmkal: Tried it out in https://github.com/mmkal/expect-type/pull/111 - can leave as a draft for now, I don't feel super strongly about which approach is better, just felt that keep the package-level testing close to where we publish a candidate package seemed right. Less worry about running `pnpm build` too many/too few times etc.

But there's not much in it!

aryaemami59: I'm Not sure tbh, I've tried this approach in a bunch of other repos it hasn't given me any issues but if you think #111 is a better approach I'd say let's go for it. We can always come back to this one if that one gives us any problems.

.github/workflows/ci.yml
mmkal: I'm ok with running it twice, I think it's better than removing the build step from here.

I just pushed what I think is a compromise solution between this and #111 - closer to yours, which I think is better since it avoids a dependency on some external service. But also I didn't love how this one has to use upload-artifact and download-artifact.

Now, it leaves this job untouched, and just checks out, installs, and packs in the are-the-types-wrong job.

Tradeoffs:

- doesn't impact the main `ci` job 👍 
- doesn't upload-artifact and download-artifact 👍 
- doesn't depend on pkg.pr.new 👍 
- does a full `pnpm install` so has my made-up node_modules risk 👎 
- builds twice 👎 
- runs in parallel and GitHub is paying for it to build twice anyway 👍👍👍 
- my node_modules risk is almost certainly bogus though 👍

.github/workflows/ci.yml
```diff
@@ -24,3 +24,12 @@ jobs:
...
       - run: pnpm type-check
+
+  are-the-types-wrong:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: corepack enable
+      - run: pnpm install
+      - run: pnpm pack
+      - run: pnpx @arethetypeswrong/cli@latest $(ls | grep tgz)
```
aryaemami59, line 35: We could just do:
```suggestion
      - run: pnpx @arethetypeswrong/cli@latest -P
```

.github/workflows/ci.yml
mmkal: I flip-flopped back. Closed #111: https://github.com/mmkal/expect-type/pull/108#discussion_r1722340526

.github/workflows/ci.yml
mmkal: nice!

.github/workflows/ci.yml
aryaemami59: Works for me! Does it need anything else?

mmkal (approved):
Sorry pushed something at the same time as you! I prefer longopts for CI (`--pack` instead of `-P`) because I'm forgetful!

Anyway sorry that there was so ridiculously much back and forth on this change. No good deed goes unpunished!

aryaemami59:
@mmkal No need for apologies my friend. Are you happy with the current approach?

mmkal:
> @mmkal No need for apologies my friend. Are you happy with the current approach?

Yes!

PR #113, opened 2024-08-22
idea: `.branded.inspect` to view deep prop types

mmkal:
Use `.branded.inspect` to find badly-defined paths:

This finds `any` and `never` types deep within objects. This can be useful for debugging, since you will get autocomplete for the bad paths, but is a fairly heavy operation, so use with caution for large/complex types.

```typescript
const bad = (metadata: string) => ({
  name: 'Bob',
  dob: new Date('1970-01-01'),
  meta: {
    raw: metadata,
    parsed: JSON.parse(metadata), // whoops, any!
  },
  exitCode: process.exit(), // whoops, never!
})

expectTypeOf(bad).returns.branded.inspect({
  foundProps: {
    '.meta.parsed': 'any',
    '.exitCode': 'never',
  },
})

const good = (metadata: string) => ({
  name: 'Bob',
  dob: new Date('1970-01-01'),
  meta: {
    raw: metadata,
    parsed: JSON.parse(metadata) as unknown, // here we just cast, but you should use zod/similar validation libraries
  },
  exitCode: 0,
})

expectTypeOf(good).returns.branded.inspect({
  foundProps: {},
})

expectTypeOf(good).returns.branded.inspect<{findType: 'unknown'}>({
  foundProps: {
    '.meta.parsed': 'unknown',
  },
})
```

---

PR notes: this also adds a `nominalTypes` option to `DeepBrand`. Reason being, a type like `Date` which has tons of methods can blow up the size of the type, and hurt performance/make us more likely to hit the dreaded "type instantiation is excessively deep" error. So now, by default DeepBrand is passed `nominalTypes: {Date: Date}` which basically means if it sees a type matching `Date` (using `MutuallyExtends`), it will just put `{type: 'Date'}` in that spot in the big branded schema thing, and stop recursing.

In theory people could configure this so they could do: 

```ts
expectTypeOf<MyType>()
  .branded.configure<{
    nominalTypes: {
      Date: Date,
      S3Bucket: awscdk.s3.Bucket,
    },
  }>()
  .toEqualTypeOf<{foo: awscdk.s3.Bucket}>()
```

Which should improve performance and reliability.

But mostly I just added it to make sure `Date` didn't break the new `inspect` functionality. Probably needs more careful thinking.

This could go in post-v1 since it's non-breaking.


src/branding.ts
```diff
@@ -26,60 +46,149 @@ import {
...
       : IsUnknown<T> extends true
         ? {type: 'unknown'}
-        : T extends string | number | boolean | symbol | bigint | null | undefined | void
-          ? {
-              type: 'primitive'
-              value: T
-            }
-          : T extends new (...args: any[]) => any
+        : Not<IsNever<NominalType<T, Options>>> extends true
+          ? {type: NominalType<T, Options>}
```
mmkal, line 57: I'm sure there's a better way to do this without having to explicitly check for never/user `NominalType` twice

src/messages.ts
```diff
@@ -47,7 +47,7 @@ export type MismatchInfo<Actual, Expected> =
             K extends keyof Expected ? Expected[K] : never
           >
         }
-    : StrictEqualUsingBranding<Actual, Expected> extends true
+    : StrictEqualUsingBranding<Actual, Expected, DeepBrandOptionsDefaults> extends true
```
mmkal, line 50: this is temp - we should make sure we respect whatever options the `.branded` that's calling this has mebe??

src/messages.ts
```diff
@@ -143,7 +143,10 @@ export type ExpectNever<T> = {[expectNever]: T; result: IsNever<T>}
  * @internal
  */
 const expectNullable = Symbol('expectNullable')
-export type ExpectNullable<T> = {[expectNullable]: T; result: Not<StrictEqualUsingBranding<T, NonNullable<T>>>}
+export type ExpectNullable<T> = {
+  [expectNullable]: T
+  result: Not<StrictEqualUsingBranding<T, NonNullable<T>, DeepBrandOptionsDefaults>>
```
mmkal, line 148: this is temp - we should make sure we respect whatever options the `.branded` that's calling this has mebe?? or should this not be using branding at all?

probably branding should always be opt-in since it can be slow

aryaemami59:
Seems like this would have very niche use cases. What's the difference between:

```ts
expectTypeOf(bad).returns.branded.inspect({
  foundProps: {
    '.meta.parsed': 'any',
  },
})
```

and this:

```ts
expectTypeOf(bad)
  .returns.toHaveProperty('meta')
  .toHaveProperty('parsed')
  .toBeAny()
```

mmkal:
It's basically DX. The toHaveProperty one is useful when you deliberately made meta.parsed any. The .branded.inspect one is useful if you have a big complex object and you want to make sure that there are no anys hiding in it - you don't know its meta.parsed, it could be foo.bar.baz.abc.def or whatever.

So basically I think I'd only check anything in when foundProps is empty. But when something goes wrong, it'll find where the bad types are hiding.

src/branding.ts
mmkal: 🤖 Done in 8b0934b — captured the lookup once with `NominalType<T, Options> extends infer Nominal` and reuse `Nominal` for both the `IsNever` guard and the `{type: Nominal}` result. The `IsNever` check stays (it's how we detect "no nominal match"), but it no longer re-resolves the type.

src/messages.ts
mmkal: 🤖 Fixed in 8b0934b — `MismatchInfo` now takes an `Options` param (defaulting to `DeepBrandOptionsDefaults`) and `.branded.toEqualTypeOf` passes its configured options through, so the error-hint branding check honours custom nominal types. Heads-up: in the recursive-`this` sabotage snapshot the default now prints as `MismatchInfo<this, this, DeepBrandOptionsDefaults>`; normal errors elide it, so it's cosmetic and confined to that one exotic case.

src/messages.ts
mmkal: 🤖 Went with "not using branding at all" in 8b0934b. `toBeNullable` has no branded entry point, so the `DeepBrand` walk was just a slow way to compare `T` vs `NonNullable<T>` with zero configurability benefit. Swapped it for `StrictEqualUsingTSInternalIdenticalToOperator` (the fast TS-internal path) — branding stays opt-in via `.branded` as you suggested.

mmkal:
🤖 Good call — added a `.toBeAny()` vs `.branded.inspect()` section to the README (generated from a new usage-test example, pushed in 6807276). It frames it as "answering different questions": `.toBeAny()` / `.toHaveProperty(...).toBeAny()` for when you already know the exact spot to pin (e.g. a deliberately-`any` property you want a regression test on, or asserting something is *not* `any` after typing it), and `.branded.inspect(...)` for surfacing anys hiding *anywhere* in a large/complex type when you don't know where to look. Let me know if you'd like the wording tweaked.

PR #124, opened 2024-10-09
[ci] try artifact.ci

mmkal:
artifact.ci is a side-project adapted from something I built on a previous team, which adds a clickable link for uploaded artifacts. I've isntalled it as a GitHub app for this repo, so just going to test it out here.


aryaemami59:
I am super excited for this!

PR #125, opened 2024-10-09
bigint printtype

mmkal:
CC @aryaemami59 noticed during my demo that we get not-v-useful `Expected: ..., Actual: number` messages when the error involves a bigint. Since we have official bigint support now thought that should be addressed before it's shipped.

View without whitespace: https://github.com/mmkal/expect-type/pull/125/files?w=1

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,6 +332,7 @@ type Factorize = {
   (input: bigint): bigint[]
 }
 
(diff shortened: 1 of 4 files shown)


aryaemami59:
Awesome, thanks a bunch. I honestly didn't even realize.

PR #126, opened 2024-10-09
toMatchObjectType + toExtend - replacements for toMatchTypeOf

### LINKED ISSUE #55: `.toMatchTypeOf()` does not take into account `readonly` properties (undocumented case?)

mmkal:
Closes #55 (if merged, collecting feedback first)

This introduces two new methods to replace the ever-controversial `toMatchTypeOf`, which would be deprecated by this change. It won't be removed until v2, but no plans for that any time soon.

- `toExtend` - does what `toMatchTypeOf` used to do - basically just check `Actual extends Expected`
- `toMatchObjectType` - does a `Pick<...>` of actual, so you can do precise checks on a subset of your type's keys:

```ts
type MyType = {readonly a: string; b: number; c: {some: {very: {complex: 'type'}}}; d: 'another type'}

expectTypeOf<MyType>().toMatchObjectType<{a: string; b: number}>() // fails - forgot readonly
expectTypeOf<MyType>().toMatchObjectType<{readonly a: string; b?: number}>() // passes
```

Note: `toMatchObjectType` only accepts plain object expectations - no unions, no tuples, no primitives

I suspect most instances of `.toMatchTypeOf` could be happily replaced by `.toMatchObjectType`, and the word "object" being in the name makes it a bit clearer what's happening - and it is closer to a compile-time version of the jest `.toMatchObject` method.

In the remainder of cases, people can reach for `toExtend` - though this one should be more rarely used, because `extends` is a fairly weak assertion (and, in fact, you can achieve something similar with `{} as MyType satisfies {a: string}`, without using any library.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -24,7 +24,7 @@ import type {
   OverloadReturnTypes,
   OverloadsNarrowedByParameters,
 } from './overloads'
-import type {AValue, Extends, MismatchArgs, StrictEqualUsingTSInternalIdenticalToOperator} from './utils'
+import type {AValue, Extends, IsUnion, MismatchArgs, Not, StrictEqualUsingTSInternalIdenticalToOperator} from './utils'
 
 export * from './branding' // backcompat, consider removing in next major version
 export * from './messages' // backcompat, consider removing in next major version
@@ -36,6 +36,28 @@ export * from './utils' // backcompat, consider removing in next major version
  * {@linkcode expectTypeOf()} utility.
  */
 export interface PositiveExpectTypeOf<Actual> extends BaseExpectTypeOf<Actual, {positive: true; branded: false}> {
+  toMatchObjectType: <
+    Expected extends IsUnion<Expected> extends true
+      ? 'toMatchObject does not support union types'
+      : Not<Extends<Expected, Record<string, unknown>>> extends true
+        ? 'toMatchObject only supports object types'
+        : StrictEqualUsingTSInternalIdenticalToOperator<
+              Pick<Actual, keyof Actual & keyof Expected>,
+              Expected
+            > extends true
+          ? unknown
+          : MismatchInfo<Pick<Actual, keyof Actual & keyof Expected>, Expected>,
+  >(
+    ...MISMATCH: MismatchArgs<
+      StrictEqualUsingTSInternalIdenticalToOperator<Pick<Actual, keyof Actual & keyof Expected>, Expected>,
+      true
+    >
+  ) => true
+
+  toExtend<Expected extends Extends<Actual, Expected> extends true ? unknown : MismatchInfo<Actual, Expected>>(
+    ...MISMATCH: MismatchArgs<Extends<Actual, Expected>, true>
+  ): true
+
   toEqualTypeOf: {
     /**
      * Uses TypeScript's internal technique to check for type "identicalness".
@@ -118,8 +140,20 @@ export interface PositiveExpectTypeOf<Actual> extends BaseExpectTypeOf<Actual, {
     ): true
   }
 
+  toExtend: <Expected extends Extends<Actual, Expected> extends true ? unknown : MismatchInfo<Actual, Expected>>(
+    ...MISMATCH: MismatchArgs<Extends<Actual, Expected>, true>
+  ) => true
+
+  /**
+   * @deprecated - use either `toMatchObject` or `toExtend` instead
+   * - use `toMatchObject` to performa a strict check on a subset of your type's keys
+   * - use `toExtend` to check if your type extends the expected type
+   */
   toMatchTypeOf: {
     /**
+     * @deprecated - use either `toMatchObject` or `toExtend` instead
+     * - use `toMatchObject` to performa a strict check on a subset of your type's keys
+     * - use `toExtend` to check if your type extends the expected type
      * A less strict version of {@linkcode toEqualTypeOf | .toEqualTypeOf()}
      * that allows for extra properties.
      * This is roughly equivalent to an `extends` constraint
@@ -147,6 +181,9 @@ export interface PositiveExpectTypeOf<Actual> extends BaseExpectTypeOf<Actual, {
     ): true
 
     /**
+     * @deprecated - use either `toMatchObject` or `toExtend` instead
+     * - use `toMatchObject` to performa a strict check on a subset of your type's keys
+     * - use `toExtend` to check if your type extends the expected type
      * A less strict version of {@linkcode toEqualTypeOf | .toEqualTypeOf()}
      * that allows for extra properties.
      * This is roughly equivalent to an `extends` constraint
@@ -265,6 +302,15 @@ export interface PositiveExpectTypeOf<Actual> extends BaseExpectTypeOf<Actual, {
  * Represents the negative expectation type for the {@linkcode Actual} type.
  */
 export interface NegativeExpectTypeOf<Actual> extends BaseExpectTypeOf<Actual, {positive: false}> {
+  toMatchObjectType: <Expected>(
+    ...MISMATCH: MismatchArgs<
+      StrictEqualUsingTSInternalIdenticalToOperator<Pick<Actual, keyof Actual & keyof Expected>, Expected>,
+      false
+    >
+  ) => true
+
+  toExtend<Expected>(...MISMATCH: MismatchArgs<Extends<Actual, Expected>, false>): true
+
   toEqualTypeOf: {
     /**
      * Uses TypeScript's internal technique to check for type "identicalness".
@@ -933,6 +979,8 @@ export const expectTypeOf: _ExpectTypeOf = <Actual>(
     toMatchTypeOf: fn,
     toEqualTypeOf: fn,
     toBeConstructibleWith: fn,
+    toMatchObjectType: fn,
+    toExtend: fn,
     toBeCallableWith: expectTypeOf,
     extract: expectTypeOf,
     exclude: expectTypeOf,
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -227,3 +227,7 @@ export type TuplifyUnion<Union, LastElement = LastOf<Union>> =
  * Convert a union like `1 | 2 | 3` to a tuple like `[1, 2, 3]`.
  */
 export type UnionToTuple<Union> = TuplifyUnion<Union>
+
+export type IsTuple<T> = Or<[Extends<T, []>, Extends<T, [any, ...any[]]>]>
+
+export type IsUnion<T> = Not<Extends<UnionToTuple<T>['length'], 1>>
--- a/test/__snapshots__/errors.test.ts.snap
+++ b/test/__snapshots__/errors.test.ts.snap
@@ -6,6 +6,11 @@ exports[`usage.test.ts 1`] = `
 
 999   expectTypeOf({a: 1, b: 1}).toEqualTypeOf<{a: number}>()
                                                ~~~~~~~~~~~
+test/usage.test.ts:999:999 - error TS2344: Type '{ b: number; }' does not satisfy the constraint '{ a: "Expected: never, Actual: number"; b: "Expected: number, Actual: never"; }'.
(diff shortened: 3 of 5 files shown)


mrazauskas:
I see `b?: number` as problem.  Imagine having a type with `b?: number` and updating it to `b: number`. I would like to see a failing test just like with:

```ts
expectTypeOf<Pick<MyType, "a" | "b">>().toEqualTypeOf<{readonly a: string; b?: number}>() // fail
```

---

Also not sure if this was intended, because documentation says: "This is a strict check, but only on the subset of keys that are in both types." Or I missed something?

mmkal:
Yep, optional checks are supported too. All the goodness from `toEqualTypeOf`, it's basically equivalent to manually doing `Pick`. Have updated the comment in the body. What do you think the docs should say to explain it better?

Also 

```ts
type MyType = {readonly a: string; b: number; c: {some: {very: {complex: 'type'}}}; d?: boolean}

expectTypeOf<MyType>().toMatchObjectType<{a: string; b: number}>() // fails - forgot readonly
expectTypeOf<MyType>().toMatchObjectType<{readonly a: string; b?: number}>() // fails - b shouldn't be optional
expectTypeOf<MyType>().toMatchObjectType<{readonly a: string; d: boolean}>() // fails - d should be optional

expectTypeOf<MyType>().toMatchObjectType<{readonly a: string; b: number}>() // passes
expectTypeOf<MyType>().toMatchObjectType<{readonly a: string; d?: boolean}>() // passes

type BinaryOp = {
  (a: number, b: number): number
  (a: bigint, b: bigint): bigint
}

type Calculator = {add: BinaryOp; subtract: BinaryOp}

expectTypeOf<Calculator>().toMatchObjectType<{add: BinaryOp}>()
expectTypeOf<Calculator>().toMatchObjectType<{subtract: BinaryOp}>()
expectTypeOf<Calculator>().toMatchObjectType<{add: BinaryOp; subtract: BinaryOp}>()

expectTypeOf<Calculator>().toMatchObjectType<{add: {(a: number, b: number): number; (a: bigint, b: bigint): bigint}}>()

expectTypeOf<Calculator>().toMatchObjectType<{add: (a: number, b: number) => number}>() // fails - only one overload
expectTypeOf<Calculator>().toMatchObjectType<{add: (a: bigint, b: bigint) => bigint}>() // fails - only one overload
```

(also, noticed you're working on something similar https://github.com/tstyche/tstyche - looks nice! have you had a similar decision to make there?)

mrazauskas:
Ah.. There was a typo. Got it. As you noticed somewhere: it is hard to explain what 'match' means in type context. For me it feels like `Pick` is familiar and clear. I was simply wondering about your opinion.

(Yes, I had `.toMatch` in TSTyche. It is deprecated in favour of `Pick` / `Omit`. By the way, only the surface looks similar. Yes, `expect()` API is somewhat familiar, but types are compared programatically like in `tsd`. Only that the install size of TSTyche is just above 220kB.)

aryaemami59:
@mmkal Can `.toExtend()` be negated?

aryaemami59:
Is `.toExtend()` basically the same as `.toMatchTypeOf()`? Is it an attempt to rename it to something less confusing or is it stricter as well?

mmkal:
> Is `.toExtend()` basically the same as `.toMatchTypeOf()`? Is it an attempt to rename it to something less confusing or is it stricter as well?

Yes

src/index.ts
```diff
@@ -118,8 +140,20 @@ export interface PositiveExpectTypeOf<Actual> extends BaseExpectTypeOf<Actual, {
     ): true
   }
 
+  toExtend: <Expected extends Extends<Actual, Expected> extends true ? unknown : MismatchInfo<Actual, Expected>>(
+    ...MISMATCH: MismatchArgs<Extends<Actual, Expected>, true>
+  ) => true
+
+  /**
+   * @deprecated - use either `toMatchObject` or `toExtend` instead
+   * - use `toMatchObject` to performa a strict check on a subset of your type's keys
```
aryaemami59, line 149:
```suggestion
   * @deprecated - use either `toMatchObjectType` or `toExtend` instead
   * - use `toMatchObjectType` to perform a strict check on a subset of your type's keys
```

src/index.ts
```diff
@@ -118,8 +140,20 @@ export interface PositiveExpectTypeOf<Actual> extends BaseExpectTypeOf<Actual, {
...
+   * - use `toExtend` to check if your type extends the expected type
+   */
   toMatchTypeOf: {
     /**
+     * @deprecated - use either `toMatchObject` or `toExtend` instead
+     * - use `toMatchObject` to performa a strict check on a subset of your type's keys
```
aryaemami59, line 155:
```suggestion
     * @deprecated - use either `toMatchObjectType` or `toExtend` instead
     * - use `toMatchObjectType` to perform a strict check on a subset of your type's keys
```

src/index.ts
```diff
@@ -147,6 +181,9 @@ export interface PositiveExpectTypeOf<Actual> extends BaseExpectTypeOf<Actual, {
     ): true
 
     /**
+     * @deprecated - use either `toMatchObject` or `toExtend` instead
+     * - use `toMatchObject` to performa a strict check on a subset of your type's keys
```
aryaemami59, line 185:
```suggestion
     * @deprecated - use either `toMatchObjectType` or `toExtend` instead
     * - use `toMatchObjectType` to perform a strict check on a subset of your type's keys
```

README.md
```diff
@@ -89,31 +89,46 @@ expectTypeOf({a: 1}).toEqualTypeOf({a: 2})
...
 type Apple = {type: 'Fruit'; name: 'Apple'; edible: true}
 
-expectTypeOf<Apple>().toMatchTypeOf<Fruit>()
+expectTypeOf<Apple>().toExtend<Fruit>()
 
 // @ts-expect-error
-expectTypeOf<Fruit>().toMatchTypeOf<Apple>()
+expectTypeOf<Fruit>().toExtend<Apple>()
+
+// @ts-expect-error - edible:boolean !== edible:true
```
aryaemami59, line 130: I think we should elaborate on `edible:boolean !== edible:true`

README.md
```diff
@@ -89,31 +89,46 @@ expectTypeOf({a: 1}).toEqualTypeOf({a: 2})
...
 
-`.toEqualTypeOf` and `.toMatchTypeOf` both fail on missing properties:
+To check that a type extends another type, use `.toExtend`:
+
+```typescript
+expectTypeOf<string>().toExtend<string | boolean>()
+// @ts-expect-error
+expectTypeOf<{a: number}>().toExtend<{b: number}>()
+
+expectTypeOf<{a: number; b: number}>().toExtend<{a: number}>() // this works, but you would be better off using `.toMatchObjectType` here
```
aryaemami59, line 105: In what way would they be better off using `.toMatchObjectType` in this situation?

test/deprecations.test.ts
```diff
@@ -0,0 +1,11 @@
+import * as fs from 'node:fs'
+import * as path from 'node:path'
+import {test, expect} from 'vitest'
+import {tsFileErrors} from './ts-output'
+
+test.each(['usage.test.ts', 'types.test.ts'])('%s: toMatchTypeOf matches toExtend behaviour', file => {
```
aryaemami59, line 6: Minor nit: can we actually use [`test.for`](https://vitest.dev/api/#test-for) with a `TestContext`?

aryaemami59, line 6: Minor nit: can we actually use [test.for](https://vitest.dev/api/#test-for) with a TestContext?

```suggestion
test.for(['usage.test.ts', 'types.test.ts'])('%s: toMatchTypeOf matches toExtend behaviour', (file, { expect }) => {
```

mmkal: Ah, didn't know about that. I guess we could but I'd rather not just because it's less familiar, to me at least - it relies knowing more about how vitest works. (Also I think your suggestion, isn't right, you'd need to do `([file], {expect})` rather than `(file, {expect})`, it looks like from the docs?

mmkal:
Update - it's now a "deep" pick, to match jest's toMatchObject. That enables this kind of thing:

```ts
const user = {
  email: 'a@b.com',
  name: 'John Doe',
  address: {street: '123 2nd St', city: 'New York', zip: '10001', state: 'NY', country: 'USA'},
}

expectTypeOf(user).toMatchObjectType<{name: string; address: {city: string}}>()
```

It also differentiates it vs a simple usage of `Pick<...>` - that would only work for top-level properties.

README.md
```diff
@@ -89,40 +89,73 @@ expectTypeOf({a: 1}).toEqualTypeOf({a: 2})
...
+```
+
+`.toExtend` can be used with object types, but `.toMatchObjectType` is usually a better choice when dealing with objects, since it's stricter:
+
+```typescript
+expectTypeOf({a: 1, b: 2}).toExtend<{a: number}>() // avoid this
+expectTypeOf({a: 1, b: 2}).toMatchObjectType<{a: number}>() // prefer this
+```
+
+`.toEqualTypeOf`, `.toMatchObjectType`, and `.toExtend` all fail on missing properties:
```
mrazauskas, line 125: As well as missing optional properties? (No time to try out. Sorry.)

Also I was wondering, why you don’t use `.not` in these examples? Because `// @ts-expect-error` makes these assertions pass also with older versions of this library.

mmkal: >As well as missing optional properties? (No time to try out. Sorry.)

`toEqualTypeOf` and `toMatchObjectType` fail on missing optional properties, but `toExtend` doesn't. Added some docs + tests to cover this explicitly though.

>why you don’t use .not in these examples

Just because these examples are essentially the docs and `.not` hasn't been introduced yet. I'd be open to reordering stuff to put `.not` higher up, you raise a good point - it's less prone to "wrong" errors.

mrazauskas: To be honest, it was my stupid mistake. I just copied this code to TS Playground and tried to mess up something. All worked really really well. Until.. I realised that this was because of `// @ts-expect-error`..

mmkal:
@aryaemami59 I want to go ahead with this one too. Removing `.toMatchTypeOf` would be a big change, but arguably deprecating it isn't. Still, I'll do this is in its own release.

Open to eleventh hour "don't do this" thoughts though if you have them!

aryaemami59 (changes requested):

aryaemami59 (approved):
Everything looks great, Since `.toMatchTypeOf` was widely used I think I might write a codemod for this.

mmkal:
> Everything looks great, Since `.toMatchTypeOf` was widely used I think I might write a codemod for this.

Worth a try, but it will need type awareness. From auditing trpc's many usages, the correct updates to be made are:

- some cases of toEqualTypeOf, where the types are identical and a much stronger assertion is beneficial
- some cases of toMatchObjectType where a somewhat stricter assertion is possible 
- toExtend for everything else

You *could* write a codemod that was simpler and replaces all instances with toExtend but it would be slightly defeating the point of the change IMO.

PR #129, opened 2024-11-20
chore: add `.gitattributes`

aryaemami59:


mmkal (approved):
I keep going back and forth on whether this is necessary, but I think that's just a function of whether projects I work on have contributors who use Windows. Happy to try this, I'm on mac anyway so I think this does no harm (to me).

PR #130, opened 2024-12-04
ci: fix TypeScript v5.0 installation issue during CI

aryaemami59:
## Overview

During CI, the `test-types` job encounters an issue where specifying `typescript@5.0` results in the `.0` being truncated. Consequently, `pnpm` installs `typescript@v5`, which resolves to the latest version (currently `v5.6.2`) instead of the intended `5.0.4`.

---

## **This PR**:

- [X] Wraps the TypeScript version numbers in strings to resolve the issue.


mmkal (approved):
Good catch. Yaml sux :(

PR #142, opened 2025-02-14
ci: fix `html` reporter in CI

aryaemami59:
##**This PR**:

- [X] Fixes `html` reporter in CI.


mmkal (approved):
nice, is not needing `--` new in pnpm 10?

aryaemami59:
> nice, is not needing `--` new in pnpm 10?

yeah we don't need it anymore.

PR #189, opened 2025-12-09
Implement an explicit error message

charpeni:
## Summary

Improve error messages for `toBeAny()`, `toBeString()`, `toBeNumber()`, and other type assertion methods to clearly show expected vs actual types

Fixes #179

### Before

```
error TS2349: This expression is not callable.
  Type 'ExpectAny<2>' has no call signatures.
```

### After

```
error TS2349: This expression is not callable.
  Type '{ "Expected: any, Actual: literal number: 2": never; }' has no call signatures.
```

For negated assertions:

```
error TS2349: This expression is not callable.
  Type '{ ".not.toBeAny() failed; Actual: any": never; }' has no call signatures.
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -564,72 +564,72 @@ export interface BaseExpectTypeOf<Actual, Options extends {positive: boolean}> {
   /**
    * Checks whether the type of the value is `any`.
    */
-  toBeAny: Scolder<ExpectAny<Actual>, Options>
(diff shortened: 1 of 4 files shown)


mmkal:
I'm going to close this as no response for a while and I'm trying to get PRs under control. With LLMs being pretty trustworthy when requirements are well-specced, I would rather prompt the implementation for this myself, so if you still want this feature, please open an issue with some good examples of how you think the lib should behave (more or less what you have in the PR body would probably be fine).

PR #203, opened 2026-03-12
refactor!: rename `_ExpectTypeOf` exported type

### LINKED ISSUE #106: Rename the `_ExpectTypeOf` type
If `_ExpectTypeOf` is going to be exported, we should probably rename it to something else to avoid confusion between `_ExpectTypeOf` and `ExpectTypeOf`.

aryaemami59:
# Summary

- Renames the exported type **`_ExpectTypeOf`** to **`ExpectTypeOfFunction`** to avoid confusion with **`ExpectTypeOf`** and follow conventional public API naming (no leading underscore).
- Resolves #106.

> [!CAUTION]
> Any consumers importing **`_ExpectTypeOf`** directly will need to update to the new name.

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -1,0 +1,2 @@
+onlyBuiltDependencies:
+  - esbuild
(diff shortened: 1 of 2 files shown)


mmkal:
If we ever release a v2 we should consider this but since this is a type-only library we should consider renames like this as a breaking change. I don't think this change justifies v2 by itself, so closing for now.

